PR #10, opened 2021-07-12
feat(components): add avatar component

### LINKED ISSUE #7: Implement Avatar Component
TODOs:

- [x] - Add specification
- [x] - Implement Component
- [ ] - Add typedoc

SisIvanova:
Closes #7

### DIFF
--- a/src/components/avatar/avatar.material.scss
+++ b/src/components/avatar/avatar.material.scss
@@ -1,0 +1,74 @@
+@use '../../styles/common/component';
+@use '../../styles/utilities' as *;
+
+:host {
+  display: block;
+}
+
+[part="native"] {
+  position: relative;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  user-select: none;
+  color: var(--igc-gray-800);
+  background: var(--igc-gray-400);
+  vertical-align: middle;
+  border-radius: 0;
+  outline-style: none;
+  flex-shrink: 0;
+}
+
+.rounded {
+  border-radius: calc(var(--igc-border-radius) * 4);
+}
+
+.circle {
+  border-radius: calc(var(--igc-border-radius) * 22);
+}
+
+img {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  border-radius: inherit;
+  background-color: var(--igc-gray-400);
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+[part="initials"] {
+  text-transform: uppercase;
+}
+
+.large {
+  width: var(--size, 88px);
+  height: var(--size, 88px);
+
+  [part="initials"] {
+    font-size: rem(44px);
+    line-height: rem(44px);
+  }
+}
+
+.medium {
+  width: var(--size, 64px);
+  height: var(--size, 64px);
+
+  [part="initials"] {
+    font-size: rem(32px);
+    line-height: rem(32px);
+  }
+}
+
+.small {
+  width: var(--size, 40px);
+  height: var(--size, 40px);
+
+  [part="initials"] {
+    font-size: rem(20px);
+    line-height: rem(20px);
+  }
+}
+
--- a/src/components/avatar/avatar.material.scss
+++ b/src/components/avatar/avatar.material.scss
@@ -2,7 +2,7 @@
 @use '../../styles/utilities' as *;
 
 :host {
-  display: block;
+  display: inline-block;
 }
 
 [part="native"] {
@@ -13,7 +13,6 @@
   user-select: none;
   color: var(--igc-gray-800);
   background: var(--igc-gray-400);
-  vertical-align: middle;
   border-radius: 0;
   outline-style: none;
   flex-shrink: 0;
@@ -32,10 +31,8 @@ img {
   width: 100%;
   height: 100%;
   border-radius: inherit;
-  background-color: var(--igc-gray-400);
-  background-size: cover;
-  background-repeat: no-repeat;
-  background-position: center;
+  background-color: inherit;
+  object-fit: cover;
 }
 
 [part="initials"] {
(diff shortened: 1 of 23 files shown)


src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
+@use '../../styles/common/component';
+@use '../../styles/utilities' as *;
+
+:host {
+  display: block;
```
simeonoff, line 5: This should rather be `inline-block` than `block` as we would want the element to appear alongside other comopnents/elements.

src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
...
+
+[part="native"] {
+  position: relative;
+  display: inline-flex;
+  justify-content: center;
+  align-items: center;
+  user-select: none;
+  color: var(--igc-gray-800);
+  background: var(--igc-gray-400);
+  vertical-align: middle;
```
simeonoff, line 16: You don't need this line.

src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
...
+
+img {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  border-radius: inherit;
+  background-color: var(--igc-gray-400);
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
```
simeonoff, line 38:
```suggestion
```

src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
...
+  border-radius: calc(var(--igc-border-radius) * 22);
+}
+
+img {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  border-radius: inherit;
+  background-color: var(--igc-gray-400);
+  background-size: cover;
```
simeonoff, line 36:
```suggestion
  object-fit: cover;
```

src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
...
+.circle {
+  border-radius: calc(var(--igc-border-radius) * 22);
+}
+
+img {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  border-radius: inherit;
+  background-color: var(--igc-gray-400);
```
simeonoff, line 35:
```suggestion
  background-color: inherit;
```

src/components/avatar/avatar.material.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+.large {
+  width: var(--size, 88px);
+  height: var(--size, 88px);
+
+  [part="initials"] {
+    font-size: rem(44px);
+    line-height: rem(44px);
+  }
```
simeonoff, line 52:
```suggestion
  $size: var(--size, rem(88px));
  width: $size;
  height: $isze;

  [part="initials"] {
    font-size: calc($size / 2);
    line-height: calc($size / 2);
  }
```

Follow the same approach for the medium and small sizes.

simeonoff (commented):
Add the ability to modify the alt text and the image `src` in the Avatar story.

src/components/avatar/avatar.ts
```diff
@@ -0,0 +1,73 @@
...
+  @state()
+  private hasError = false;
+
+  @property()
+  alt?: string;
+
+  @property()
+  initials?: string;
+
+  @property()
```
simeonoff, line 23:
```suggestion
  @property({reflect: true})
```

test/avatar.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  it('should set avatar size correctly', async () => {
+    const el = await fixture<IgcAvatarComponent>(
+      html`<igc-avatar size="small"></igc-avatar>`
+    );
+
+    expect(el.size).to.equal('small');
+    el.setAttribute('size', 'medium');
+    expect(el).dom.to.equal(`<igc-avatar size="medium"></igc-avatar>`);
```
simeonoff, line 30: You should be checking the instance value of size, rather than whether the setAttribute works as expected.

test/avatar.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  it('should set avatar shape correctly', async () => {
+    const el = await fixture<IgcAvatarComponent>(
+      html`<igc-avatar shape="square"></igc-avatar>`
+    );
+
+    expect(el.shape).to.equal('square');
+
+    el.setAttribute('shape', 'rounded');
+    expect(el).dom.to.equal(`<igc-avatar shape="rounded"></igc-avatar>`);
```
simeonoff, line 41: Same as above.

simeonoff (dismissed):

rkaraivanov (dismissed):

src/components/avatar/avatar.ts
```diff
@@ -0,0 +1,76 @@
...
+import { property, state } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { Constructor } from '../common/mixins/constructor.js';
+import { SizableMixin } from '../common/mixins/sizable.js';
+import { styles } from './avatar.material.css';
+
+export class IgcAvatarComponent extends SizableMixin(
+  <Constructor<LitElement>>LitElement
+) {
```
DiyanDimitrov, line 11: This could be simplified.
```suggestion
export class IgcAvatarComponent extends SizableMixin(LitElement) {
```

src/components/avatar/avatar.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  constructor() {
+    super();
+    this.size = 'small';
+  }
+
+  render() {
+    return html`
+      <div
+        part="native"
```
DiyanDimitrov, line 50: This part should be called `base` according to the spec.

src/components/avatar/avatar.ts
```diff
@@ -0,0 +1,76 @@
...
+        aria-label="avatar"
+        aria-roledescription=${this.size + ' ' + this.shape}
+        class=${classMap(this.classes)}
+      >
+        ${this.initials
+          ? html`<span part="initials">${this.initials}</span>`
+          : html`
+              <span part="icon">
+                <slot name="icon"></slot>
+              </span>
```
DiyanDimitrov, line 61: We could use the default slot here.
@simeonoff, What do you think?

simeonoff: I agree.

src/components/avatar/avatar.ts
simeonoff: I would actually change the spec to be in line with all other components. The button and the badge now use `native` and the avatar shouldn't be the odd on in the bunch.

DiyanDimitrov: If we want to be consistent across all components I think we should use `base` instead of `native` since it describes better that this is the base container element.

simeonoff: True, fine then. Let's merge it like it is now, then we'll replace them all in a single PR.

DiyanDimitrov (approved):

simeonoff (approved):

PR #26, opened 2021-07-21
feat(radio): add implementation

simeonoff:


src/components/radio-group/radio-group.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  private _cachedRadios!: IgcRadioComponent[];
+
+  private get _slottedRadios() {
+    const slot = this.shadowRoot!.querySelector('slot');
+    const childNodes = slot!.assignedNodes({ flatten: true });
+    return Array.prototype.filter.call(childNodes, (node) => {
+      return node.nodeName.toLowerCase() == 'igc-radio';
+    }) as IgcRadioComponent[];
+  }
```
rkaraivanov, line 18:
```suggestion
@queryAssignedNodes(undefined, true, 'igc-radio')
  _slottedRadios!: NodeListOf<IgcRadioComponent>;
```

I think this should also do the trick

src/components/radio/radio.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  getSiblings() {
+    const group = this.closest('igc-radio-group');
+
+    if (!group) {
+      return [];
+    }
+
+    return ([...group.querySelectorAll('igc-radio')] as this[]).filter(
+      (radio: this) => radio.name === this.name && radio !== this
```
rkaraivanov, line 108:
```suggestion
return Array.from<IgcRadioComponent>(group.querySelectorAll('igc-radio'))
      .filter((radio) => radio.name === this.name && radio !== this);
```
IMO a little bit more clearer than the de-structuring and awkward casts.

rkaraivanov (changes requested):

src/components/radio/radio.ts
```diff
@@ -0,0 +1,140 @@
...
+export class IgcRadioComponent extends EventEmitterMixin<
+  IgcRadioEventMap,
+  Constructor<LitElement>
+>(LitElement) {
+  static styles = styles;
+
+  private inputId = `radio-${nextId++}`;
+  private labelId = `radio-label-${this.inputId}`;
+
+  @query('input[type="radio"]')
```
DiyanDimitrov, line 27: This could be cached.
```suggestion
  @query('input[type="radio"]', true)
```

src/components/radio/radio.ts
```diff
@@ -0,0 +1,140 @@
...
+        @mousedown="${this.handleMouseDown}"
+      >
+        <input
+          id="${this.inputId}"
+          type="radio"
+          name="${ifDefined(this.name)}"
+          value="${ifDefined(this.value)}"
+          .disabled="${this.disabled}"
+          .checked="${live(this.checked)}"
+          tabindex=${this.checked ? 0 : -1}
```
DiyanDimitrov, line 123: If there is no checked radio initially, we should make the first radio in a group focusable.

stories/radio-group.stories.ts
```diff
@@ -0,0 +1,76 @@
...
+const Template: Story<ArgTypes, Context> = (
+  { alignment, labelPosition, disabled }: ArgTypes,
+  { globals: { direction } }: Context
+) => {
+  return html`
+    <igc-radio-group
+      dir="${ifDefined(direction)}"
+      alignment="${ifDefined(alignment)}"
+      label-position="${ifDefined(labelPosition)}"
+      .disabled="${ifDefined(disabled)}"
```
DiyanDimitrov, line 63: `ifDefined` has no effect here since it applies only for attributes.

stories/radio.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const Template: Story<ArgTypes, Context> = (
+  { label, labelPosition, checked, disabled }: ArgTypes,
+  { globals: { direction } }: Context
+) => html`
+  <igc-radio
+    label-position="${ifDefined(labelPosition)}"
+    .disabled="${ifDefined(disabled)}"
+    .checked="${ifDefined(checked)}"
```
DiyanDimitrov, line 63: `ifDefined` has no effect here since it applies only for attributes.

src/components/radio-group/radio-group.ts
```diff
@@ -0,0 +1,86 @@
...
+  updateLabelPosition() {
+    this._slottedRadios.forEach(
+      (radio) => (radio.labelPosition = this.labelPosition)
+    );
+  }
+
+  @watch('disabled', { waitUntilFirstUpdate: true })
+  updateDisabled() {
+    if (!this._cachedRadios) this._cachedRadios = this.radios;
+    this._cachedRadios.forEach((radio) => (radio.disabled = this.disabled));
```
DiyanDimitrov, line 48: I understand that you want to modify only the radios that initially are not disabled, but what if radio buttons are added/removed at runtime?
This makes me think what we should listen for `slotchange` and modify the newly added radios' labelPosition and disabled properties. I think that we should either override all radios' properties or remove these properties from the radio group component.

stories/radio-group.stories.ts
```diff
@@ -0,0 +1,76 @@
...
+    >
+      <igc-radio
+        name="fruit"
+        value="apple"
+      >Apple</igc-radio>
+      <igc-radio name="fruit" value="orange" checked>Orange</igc-radio>
+      <igc-radio name="fruit" value="mango">Mango</igc-radio>
+      <igc-radio name="fruit" value="banana" disabled>Banana</igc-radio>
+    </igc-radio>
+`;
```
DiyanDimitrov, line 73: Missing `igc-radio-group` closing tag.

DiyanDimitrov (changes requested):

src/components/radio-group/radio-group.ts
simeonoff: I am on the side of overriding the values specified by individual radio buttons. So we should drop all caching that I am currently doing for all radio buttons in a group. Do you think we should do this on first update as well?

DiyanDimitrov: I am more on the side of removing these props from the radio group :)
@rkaraivanov, What do you think?
If we leave them I think they should definitely work on the first update which may cause overriding the individual radio's values.

src/components/radio/radio.ts
DiyanDimitrov: It is better now but there is one minor difference with the standard radio behavior. If no radio is selected and you focus the first one then press arrow down, the second radio should be selected (in our implementation the first one is selected).

src/components/radio-group/radio-group.ts
simeonoff: We have to keep parity with igniteui-angular, this is the only reason they're there.

src/components/radio/radio.ts
simeonoff: This behavior just adds unnecessary complexity to the implementation, imo. If you really wanna have it that way - fine, but I think we can do without it.

DiyanDimitrov: I am ok to leave it like this and fix in the future if a customer complains.

rkaraivanov (dismissed):

DiyanDimitrov (dismissed):

DiyanDimitrov (approved):

PR #27, opened 2021-07-22
refactor(button): add fluent and indigo themes

### LINKED ISSUE #25: Button: Add Fluent and Indigo Themes

desig9stein:
closes #25


src/styles/themes/material.scss
```diff
@@ -16,7 +16,7 @@ $elevations: utils.elevations(
 $type-scale: utils.type-scale();
 
 :root {
-  --igc-border-radius: #{utils.rem(2px)};
+  --igc-border-radius: #{utils.rem(4px)};
```
simeonoff, line 19: This PR shouldn't be changing the default global radius of the material theme.

src/components/button/button.indigo.scss
```diff
@@ -0,0 +1,169 @@
+@use '../../styles/utilities' as utils;
+
+@mixin theme() {
+	igc-button,
+	igc-link-button {
+		--l-size: #{utils.rem(36px)};
+		--m-size: #{utils.rem(32px)};
+		--s-size: #{utils.rem(24px)};
```
simeonoff, line 8: You should not be introducing new CSS variables specific to a given theme.

src/components/button/button.fluent.scss
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        &:focus,
+        &:active {
+            color: #{utils.color(primary, 500)};
+
+            &::before {
+                opacity: .8;
+                background: #{utils.color(gray, 100)};
```
simeonoff, line 48: Is there any reason not to apply the opacity directly to the color as the third parameter?

simeonoff (commented):
Good job! Let's use the opportunity to make the Fluent theme closer to the actual [Fluent UI Button](https://developer.microsoft.com/en-us/fluentui#/controls/web/button) implementation.

src/components/button/button.fluent.scss
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        &:focus,
+        &:active {
+            color: #{utils.color(gray, 900)};
+
+            &::before {
+                opacity: 1;
```
simeonoff, line 73: I believe these are no longer needed. Remove here and throughout.

desig9stein: That opacity is required in order to override the one that comes from the material theme, otherwise, it's impossible to achieve the desired look and feel.

simeonoff (changes requested):
The disabled styles of the indigo fab button differ from the non-disabled ones. Fix and we can merge.

simeonoff (approved):

PR #29, opened 2021-07-23
Add stylelint and fix Sass linting errors

simeonoff:


desig9stein (dismissed):

DiyanDimitrov (commented):
If we should have 4 spaces indentation for scss I would suggest to add this to the .editorconfig file:
```
[*.scss]
indent_size = 4
```

simeonoff:
> If we should have 4 spaces indentation for scss I would suggest to add this to the .editorconfig file:
> 
> ```
> [*.scss]
> indent_size = 4
> ```

Done.

desig9stein (approved):

DiyanDimitrov (approved):

PR #58, opened 2021-09-09
Ibarakov/nav drawer

onlyexeption:


simeonoff (commented):
The items should stretch across the entire width when the navigation is positioned at the top or the bottom. There shouldn't be extra margin at the end of each icon when the mini variant is used.

simeonoff (dismissed):

stories/nav-drawer.stories.ts
```diff
@@ -0,0 +1,169 @@
...
+  content1: string;
+  content2: string;
+  open: boolean;
+  pinned: boolean;
+  disabled: boolean;
+  active: boolean;
+  position: 'start' | 'end' | 'top' | 'bottom';
+}
+
+//end region
```
gedinakova, line 83: The stories' metadata is nor regenerated on build because of a typo in endregion - it should be one word with an interval after the slashes. 
This comment should be `// endregion`.

gedinakova (changes requested):

stories/nav-drawer.stories.ts
```diff
@@ -0,0 +1,169 @@
...
+        dir=${ifDefined(direction)}
+        .open=${open}
+        .pinned=${pinned}
+        .position=${position}
+      >
+        <igc-nav-drawer-header-item>Sample Drawer</igc-nav-drawer-header-item>
+
+        <igc-nav-drawer-item .disabled=${disabled}>
+          <igc-icon slot="icon" name="home"></igc-icon>
+          <span slot="text">${content1}</span>
```
gedinakova, line 140: Update this template - it seems like slot="text" does not exist on the drawer item component.

src/components/nav-drawer/nav-drawer.ts
```diff
@@ -0,0 +1,123 @@
+import { html, LitElement } from 'lit';
```
gedinakova, line 1: Add access modifiers to all API members.

src/components/nav-drawer/nav-drawer-header-item.ts
```diff
@@ -0,0 +1,17 @@
+import { html, LitElement } from 'lit';
```
gedinakova, line 1: Add access modifiers to all API members.

src/components/nav-drawer/nav-drawer-item.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @element igc-nav-drawer-item
+ *
+ * @slot - The default slot for the drawer item.
+ * @slot icon - The slot for the icon of the drawer item.
+ *
+ * @csspart base - The base wrapper of the drawer item.
+ * @csspart icon - The icon container.
+ * @csspart text - The text container.
+ */
+export class IgcNavDrawerItemComponent extends LitElement {
```
gedinakova, line 20: Add access modifiers to all API members.

gedinakova (changes requested):

gedinakova:
@onlyexeption @SisIvanova I know this is not in the spec, but how about adding an ellipsis if the item's content is too long? It looks bad right now:
image

stories/nav-drawer.stories.ts
SisIvanova: We needed a name for the slot in order to style the items properly.

gedinakova:
@SisIvanova The ellipsis mixin doesn't seem to work because the nav-drawer-items are display: flex.

gedinakova:
@onlyexeption @SisIvanova The mini div keeps showing even if there is no content for the mini slot (height=0 but width~60px).
image

gedinakova:
@onlyexeption What should happen if position is top or bottom and pinned=true? Currently, it fixes it at the top right with the height of a top/bottom positioned nav-drawer:
image

simeonoff (approved):

simeonoff:
> @onlyexeption What should happen if position is top or bottom and pinned=true? Currently, it fixes it at the top right with the height of a top/bottom positioned nav-drawer: image

IMO this should not be a part of the navigation drawer. The positioning should be handled by an external component like the `igc-app-host` component. A component like that should handle the positioning of all elements like the navigation drawer, including the navbar and bottom navigation.

gedinakova:
> > @onlyexeption What should happen if position is top or bottom and pinned=true? Currently, it fixes it at the top right with the height of a top/bottom positioned nav-drawer: image
> 
> IMO this should not be a part of the navigation drawer. The positioning should be handled by an external component like the `igc-app-host` component. A component like that should handle the positioning of all elements like the navigation drawer, including the navbar and bottom navigation.

@simeonoff Yes, sounds reasonable. We should probably take this for a broader discussion and consider such approach in a different issue. 
For this particular scenario of the nav-drawer, how about we do not pin it if position is top/bottom to avoid what happens now?

gedinakova:
> @SisIvanova The ellipsis mixin doesn't seem to work because the nav-drawer-items are display: flex. It does look better as shown below, but doesn't really display the ellipsis symbol. Is that the expected result? image

Fixed now!

gedinakova (approved):

PR #72, opened 2021-09-20
add decorators to allow for API corrections

gmurray81:
when generating the blazor API wrappers

Added decorators allow for the auto-generated blazor API to avoid collisions or reserved words that occur at that level. These decorators have no implementation since they are not designed to have runtime impact, only to allow for the generation tool to emit the correct info.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -27,6 +27,15 @@
       {
         "argsIgnorePattern": "^_"
       }
+    ],
+    "@typescript-eslint/explicit-member-accessibility": [
+      "warn",
+      {
+        "accessibility": "explicit",
+        "overrides": {
+          "constructors": "no-public"
+        }
+      }
     ]
   },
   "overrides": [
--- a/igniteui-webcomponents.ts
+++ b/igniteui-webcomponents.ts
@@ -1,6 +1,7 @@
 import { IgcAvatarComponent } from './src/components/avatar/avatar.js';
 import { IgcBadgeComponent } from './src/components/badge/badge.js';
 import { IgcButtonComponent } from './src/components/button/button.js';
+import { IgcFormComponent } from './src/components/form/form.js';
 import { IgcCardComponent } from './src/components/card/card.js';
 import { IgcCardActions } from './src/components/card/card.actions.js';
 import { IgcCardContent } from './src/components/card/card.content.js';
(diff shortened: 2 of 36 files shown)


DiyanDimitrov:
@gmurray81. Do we need these decorators if the members are private?

gmurray81:
No, @DiyanDimitrov, you should only need them for public members.
Private members wont be considered for being added to the wrapping API.

So first consider if something should be marked protected/private and prefer that if it shouldn't be used from the TS API either.

DiyanDimitrov:
@gmurray81, We have merged #71 which fixes some access modifiers. Could you please remove the decorators for the private/protected members in this PR?

gmurray81:
@DiyanDimitrov I've adjusted the decorators.

DiyanDimitrov:
@gmurray81, I am just wondering why do we still need alternate names for some private and protected members?

gmurray81:
@DiyanDimitrov The names we'll use for the events up ad the Blazor level are based on the method names that are raising the events (even though they are private)

Some of the event names collide with method names on the components:

E.g. Focus and Focus.

`@alternateName("handleFocused")` will result in the Blazor level event being called Focused instead.

So we need either the event name or the method name to differ up at the Blazor level and so need one of them to be renamed since these aren't allowed to collide for C#  up at the blazor level.
There's also a similar issue over on `IgcInputComponent`. In that case we have an event called Input, but the component up at the blazor level is also called Input (Product Owners havent decided if we will add prefixes yet to the components, we didn't for the other stuff we've released for Blazor yet), and C# disallows having an event with the same name as the class IIRC, so we need a different name there also. Also product owners may end up wanting different names for the events at the blazor level anyhow since there may be different naming expectations. I think it would be best to keep them equal for consistency though, but we need to handle collisions between the event names and the method names at least.

For the moment, I'm just focusing on everything being able to compile, and we can get feedback on the event naming after.

gmurray81:
As an alternative to having differing names on the events between TS and Blazor you could avoid having method names that collide with the event names. 

I know in a way they don't collide because you call the events "IgcFocus" and "IgcBlur" but realize I don't think we can use that naming up at the Blazor level because that prefix wont be meaningful there. And it wont make sense for events to have prefixes in a Blazor API.

gmurray81:
Note, though, that the `Input` event is still a problem, because it isn't allowed to collide with the containing class name.

That might go away if we decide to introduce a prefix to the component names up at the Blazor level, but that's still in discussion, so for now I need it to not collide in order for things to compile.

gmurray81:
adjust the decorators, rather.

gmurray81:
Reason I didn't start with decorators on the interface is that its a bunch more work to statically infer the relationship with the interface and then match the events up when every case in the repo up to that point I could basically just infer from the method name, so that's how I'm doing it for now, but will adjust if I need to once calendar is in place.

gmurray81:
I think you had another case btw, where there were multiple methods that could raise an event of the same name. but that still worked fine because they delegated to a common method that raised the event that had the right name:

e.g. if `handleDateClicked` and `handleDateKeyDown` called `handleChange` which emitted the event, that would work fine. But no need to twist the implementation to make the generation work. I can make it smarter when calendar gets checked in if you want it to emit the event from multiple places.

DiyanDimitrov (approved):

PR #76, opened 2021-09-27
adjust decorators to use event map interfaces

gmurray81:
These are changes to go along with the blazor generation switching to inspecting the event map interfaces to discover which events to generate, and renaming some of the colliding methods rather than the colliding events.


gmurray81:
note, because I can't use decorators on interface members, an alternateName is conveyed through a comment in those places, rather than a decorator.

rkaraivanov (approved):

PR #77, opened 2021-09-28
Add slider component

### LINKED ISSUE #39: Slider component
_Estimated Delivery Sprint - Sprint 4_

- [x] Research
- [x] Specification
- [x] API Documentation
- [x] Design implementation

mtsvyatkova:
Closes #39


rkaraivanov:
@mtsvyatkova 
Ok, can we take a different approach for slider component.

Can we:
* Separate the thumb logic and implementation as a separate component
* Make use of the `@watch` decorator we have in the repository instead of calling `requestUpdate`. As
a bonus it can wait for the first update paint.
* The implementation is very "Angularish" which is maybe not the correct approach in this case.

Maybe we can get together in a session and discuss the points above

src/components/slider/slider-base.ts
```diff
@@ -0,0 +1,515 @@
...
+    this.showThumbLabels();
+  };
+
+  private handleThumbPointerLeave = () => {
+    this.hideThumbLabels();
+  };
+
+  protected renderTicks() {
+    const groups = [];
+    for (let i = 0; i < this.totalTickCount(); i++) {
```
rkaraivanov, line 398:
```suggestion
    for (let i = 0, totalCount = totalTickCount(); i < totalCount; i++) {
```
A little nitpick but I don't think we need to call the method on each iteration. Or maybe cache the tick count and update it when the ticks properties change.

rkaraivanov (changes requested):
LGTM with some suggestions.

Also, add an entry into the changelog

DiyanDimitrov:
@gmurray81, Could you please take a look at the public API of the slider components and let us know if it will work ok for Blazor? I have some concerns for the range slider.  We have `lower` and `upper` value properties and a single `igcInput/igcChange` event is emitted when any of the two values is changed. Is that going to work ok for the Blazor bindings?

gmurray81:
Thanks for the heads up @DiyanDimitrov will look into it soon.

gedinakova (approved):

DiyanDimitrov:
> Thanks for the heads up @DiyanDimitrov will look into it soon.

@gmurray81, Did you get a chance to look into this?

gmurray81:
@DiyanDimitrov I think we should be able to make that work as long as the args have the values in them

gmurray81:
but I'll probably need to adjust something in the auto gen, at a guess

rkaraivanov (approved):

simeonoff (approved):

PR #78, opened 2021-09-29
build(*): Build and publish configuration

rkaraivanov:
Modified the entry point of the package.
Components are self-registering now.
Clean-up and fixes around the build process


scripts/_package.json
```diff
@@ -15,6 +16,28 @@
   "bugs": {
     "url": "https://github.com/IgniteUI/igniteui-webcomponents/issues"
   },
+  "keywords": [
+    "webcomponents",
+    "web components",
+    "UI components",
+    "avatar",
+    "badge",
+    "bottom nav",
```
gedinakova, line 25: We don't have a bottom nav yet.

gedinakova (changes requested):

scripts/_package.json
kdinev: I thought the nav drawer serves this purpose with a bottom position, since the mini template is always visible.

gedinakova: Actually, it's exactly the opposite - the mini is never shown when position is top or bottom.

simeonoff (approved):

gedinakova:
@kdinev @rkaraivanov The only thing left here is the README.md Usage section update.

PR #97, opened 2021-10-30
add icon api for blazor

gmurray81:
add convenience methods for icon registry for blazor.

### DIFF
--- a/src/components/icon/icon.ts
+++ b/src/components/icon/icon.ts
@@ -1,7 +1,7 @@
 import { html, LitElement } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
 import { unsafeSVG } from 'lit/directives/unsafe-svg.js';
-import { alternateName } from '../common/decorators';
+import { alternateName, blazorInclude } from '../common/decorators';
 import { SizableMixin } from '../common/mixins/sizable.js';
 import { styles } from './icon.material.css';
 import {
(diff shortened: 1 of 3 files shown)


src/components/icon/icon.ts
```diff
@@ -99,6 +105,18 @@ export default class IgcIconComponent extends SizableMixin(LitElement) {
   protected render() {
     return html` ${unsafeSVG(this.svg)} `;
   }
+
+  public async registerIcon(name: string, url: string, collection = 'default') {
+    await registerIcon_impl(name, url, collection);
+  }
+
+  public registerIconFromText(
```
DiyanDimitrov, line 113: We could probably make these static?

gmurray81: I made them instanced because we don't yet have the capacity to call a static method from the blazor side yet... but you are right, they would be cleaner as statics maybe... let me see what that would take.

gmurray81: I suspect I may need to make them static up at the blazor level, becuase we really need to wait for a component to have booted up to make sure we are prepared to communicate with the JS level stuff, without a bunch more wiring being added. We could make these private methods instead and just have them callable by blazor, as an alternative.

gmurray81: Or I may have the facility to have them be static down at the JS level, but instanced up at the blazor level, potentially

gmurray81: @DiyanDimitrov  I've shifted these to protected. I want to keep them instanced for now to have a solution available for the blazor wrapper without needing to add a bunch of new wiring up at that level. I don't think we need the static methods at the TS level, because at that level you can just talk to the icon registry directly. These are more support methods to make it easy to talk to the icon registry from the blazor level

DiyanDimitrov (approved):

rkaraivanov (approved):

PR #101, opened 2021-11-02
build(*): optimize build process

DiyanDimitrov:
Changes:
- removed `src` folder from the publish build
- changed `lit` to hard dependency
- added `sideEffects: false` in the package.json
- removed themes scss from compiling to ts/js
- removed non-material scss files from compiling to ts/js


scripts/build-styles.js
```diff
@@ -38,7 +38,7 @@ async function sassRender(sourceFile, templateFile, outputFile) {
 
 (async () => {
   const template = path.resolve(process.argv[1], '../styles.tmpl');
-  const paths = await globby(['src/components/**/*.scss', 'src/styles/themes/**/*.scss']);
+  const paths = await globby(['src/components/**/*.material.scss']);
```
gedinakova, line 41: Is this applicable to material only?

scripts/build.js
```diff
@@ -32,7 +31,9 @@ async function buildThemes() {
       outCss = outCss.substr(1);
     }
 
-    await writeFile(DEST_DIR(sassFile.replace(/\.scss$/, '.css')), outCss, 'utf-8');
+    const outputFile = DEST_DIR(sassFile.replace(/\.scss$/, '.css').replace('src/', ''));
```
gedinakova, line 34: How about replacing "src/styles" so we end up with a path like product/themes/theme.file?

gedinakova (changes requested):

scripts/build-styles.js
DiyanDimitrov: Yes, the other component theme files are only used in the global theme files.

scripts/build.js
DiyanDimitrov: Done.

gedinakova (approved):

gedinakova (approved):

rkaraivanov (approved):

PR #110, opened 2021-11-04
feat(calendar): add styles for disabled and special dates

### LINKED ISSUE #98: Calendar is missing disabled and special dates styles
## Description  
Calendar is missing disabled and special dates styles.

## Steps to reproduce  

1. Open the calendar.stories.ts file.
2. Change the disabledDates and specialDates like this:
```ts
  const today = new Date();
  const disabledDates: DateRangeDescriptor[] = [
    {
      type: DateRangeType.Between,
      dateRange: [
        new Date(today.getFullYear(), today.getMonth(), 1),
        new Date(today.getFullYear(), today.getMonth(), 5),
      ],
    },
  ];
  const specialDates: DateRangeDescriptor[] = [
    {
      type: DateRangeType.Between,
      dateRange: [
        new Date(today.getFullYear(), today.getMonth(), 6),
        new Date(today.getFullYear(), today.getMonth(), 10),
      ],
    },
  ];
```

## Result  
There is not visual styling for the disabled and special dates.

## Expected result  
There should be visual styling for the disabled and special dates. Here are Angular Calendar links for reference:
https://www.infragistics.com/products/ignite-ui-angular/angular/components/calendar#disabled-dates
https://www.infragistics.com/products/ignite-ui-angular/angular/components/calendar#special-dates

desig9stein:
closes #98


src/components/calendar/calendar.material.scss
```diff
@@ -5,12 +5,13 @@
 $date-size-large: rem(32px);
 $date-size-medium: rem(28px);
 $date-size-small: rem(24px);
-
+$calendar-bacground: color(surface);
```
simeonoff, line 8:
```suggestion
$calendar-background: color(surface);
```

src/components/calendar/calendar.material.scss
```diff
@@ -5,12 +5,13 @@
...
 
 :host {
     display: flex;
     flex-direction: column;
     font-family: var(--igc-font-family);
+    background: $calendar-bacground;
```
simeonoff, line 14:
```suggestion
    background: $calendar-background;
```

simeonoff (changes requested):

simeonoff (dismissed):

StefanIvanov (approved):

desig9stein:
@DiyanDimitrov @StefanIvanov  after fixing the styling I saw that I am able to navigate over the disabled dates using the keyboard. Is this on purpose? I am asking because I saw that in igniteui for angular this is not possible.

StefanIvanov (approved):

StefanIvanov:
> @DiyanDimitrov @StefanIvanov after fixing the styling I saw that I am able to navigate over the disabled dates using the keyboard. Is this on purpose? I am asking because I saw that in igniteui for angular this is not possible.

This looks like a bug to me, disabled dates should not be possible to navigate to by any means because they are disabled :)

DiyanDimitrov:
> > @DiyanDimitrov @StefanIvanov after fixing the styling I saw that I am able to navigate over the disabled dates using the keyboard. Is this on purpose? I am asking because I saw that in igniteui for angular this is not possible.
> 
> This looks like a bug to me, disabled dates should not be possible to navigate to by any means because they are disabled :)

Actually this is intended behavior in the IgcCalendar. The disabled dates can be navigated using the keyboard, but cannot be selected. When using a screen reader while navigating the disabled dates you will be notified that this date is disabled. If the disabled dates are not focusable we will just skip them without notifying the screen reader why.

DiyanDimitrov:
> > > > @DiyanDimitrov @StefanIvanov after fixing the styling I saw that I am able to navigate over the disabled dates using the keyboard. Is this on purpose? I am asking because I saw that in igniteui for angular this is not possible.
> > > 
> > > 
> > > This looks like a bug to me, disabled dates should not be possible to navigate to by any means because they are disabled :)
> > 
> > 
> > Actually this is intended behavior in the IgcCalendar. The disabled dates can be navigated using the keyboard, but cannot be selected. When using a screen reader while navigating the disabled dates you will be notified that this date is disabled. If the disabled dates are not focusable we will just skip them without notifying the screen reader why.
> 
> ОК I didn't see that use case and it is a valid one, but then how about Angular, should we suggest this as an enhancement there?

Yes, we could implement it with all other design changes that we have made in the IgcCalendar.

simeonoff (dismissed):

DiyanDimitrov:
Clicking on outside days is broken, we should switch the month to previous/next month when clicking leading/trailing days.

DiyanDimitrov:
> Clicking on outside days is broken, we should switch the month to previous/next month when clicking leading/trailing days.

It was an issue with the parts that we apply. I have fixed it.

DiyanDimitrov (approved):

PR #113, opened 2021-11-04
attributes can't be set during constructor, need to wait till connect…

gmurray81:
…edCallback at least.

### DIFF
--- a/src/components/list/list-header.ts
+++ b/src/components/list/list-header.ts
@@ -16,6 +16,10 @@ export default class IgcListHeaderComponent extends LitElement {
 
   constructor() {
     super();
+  }
+
+  public connectedCallback() {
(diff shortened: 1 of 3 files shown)


damyanpetev (approved):

gmurray81:
going to fast track this one since we need it for imminent demos, plus icon is already doing this right. Per the standard you can't add any children or attributes during the constructor. For whatever reason hydrating from markup often doesn't complain about this, but createElement or calling the constructor directly generally does. There are similar bugs in calendar that we'll want to fix, but we don't need calendar for the demo.

gmurray81:
There appears to be an issue with the Github Actions ATM. Going to stomp this one in I'm afraid. Need to get this in today.

PR #123, opened 2021-11-08
Add import eslint plugin

DiyanDimitrov:
I have been searching for an option to add the extensions of the imports during the Typescript compilation but unfortunately Typescript does not provide such option. Here is the [issue](https://github.com/microsoft/TypeScript/issues/16577) which they closed with this [comment](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937).

There are some third party solutions like this [package](https://github.com/Zoltu/typescript-transformer-append-js-extension) however I am not sure we should do something that Typescript does not want to provide.

That's why I have added an eslint plugin that will force us to have the extensions in the imports. Unfortunately we should manually fix all our imports.

@rkaraivanov @damyanpetev Please let me know what do you think on the matter.


rkaraivanov:
I am closing this issue.
There is currently some experimental support for ESM by the Typescript compiler itself.
Once [this](https://www.typescriptlang.org/docs/handbook/esm-node.html) is considered stable, we can take a look at it and update the relevant configuration.

Issue #134, opened 2021-11-11
Button border is not visible in some scenarios


igdmdimitrov:
## Description  
Button border is not visible in some scenarios

## Steps to reproduce  

1. Open button topic for Ignite UI Web Components
2. Observe the sample in outlined button section

## Result  
Border is partially not visible
![image](https://user-images.githubusercontent.com/49060557/141269554-80fc5f10-3f05-442e-9ceb-9d4e2935dfe2.png)

## Expected result  
Border should be visible

simeonoff:
After some deliberation we've decided that this should be handled on a case by case basis, i.e. the sample view should add a bit of padding to compensate for the border and focus outlines.

PR #136, opened 2021-11-12
refactor(bootstrap-theme): improve styles

desig9stein:
related to #134

1. Adding margin around all elements with outline border.
2. Fix the styles for switch element to match the bootstrap v5 styles


simeonoff:
Closing for now as this should be handled on the sample level.

PR #139, opened 2021-11-18
Rating component

### LINKED ISSUE #4: Rating component
Create an input for rating that can both show the rating of an item and let the user choose a value e.g. when leaving feedback
image

rkaraivanov:
Closes #4 #189


gedinakova (changes requested):

src/components/common/decorators/watch.ts
```diff
@@ -34,7 +34,7 @@ export function watch(propName: string, options?: WatchOptions) {
 
         if (oldValue !== newValue) {
           if (!options?.waitUntilFirstUpdate || this.hasUpdated) {
-            this[name].call(this, oldValue, newValue);
+            this[name].call(this, newValue, oldValue);
```
DiyanDimitrov, line 37: I think we should leave this as it was before since you may need only the old value, the new value could be retrieved from the property itself.

DiyanDimitrov (changes requested):
We may consider adding a property to highlight only the selected item similar to [this](https://mui.com/components/rating/#radio-group), but it is not necessarily a v1 feature.

src/components/rating/rating.ts
```diff
@@ -0,0 +1,308 @@
...
+  @property({ type: Number })
+  public precision = 1;
+
+  /**
+   * The symbol which the rating will display.
+   * It also accepts a callback function which gets the current symbol
+   * index so the symbol can be resolved per position.
+   */
+  @property()
+  public symbol: string | ((index: number) => string) = '⭐';
```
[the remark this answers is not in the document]
rkaraivanov: I'm fine with that change.
@gmurray81 Will you chime on this?

As for the _emptySymbol_ property this would entail an _emptySymbolFormatter_ as well. 
Additionally this should be
discussed with the design team on the implementation and the looks. In the case of decimal step you would have a selected part of the symbol/not a selected part of the symbol/ an empty symbol or would be selected/part empty/empty?

src/components/rating/rating.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  /**
+   * Decrements the value of the control by `n` steps multiplied by
+   * the precision factor.
+   */
+  public stepDown(n = 1) {
+    this.value -= this.round(n * this.precision);
+  }
+
+  protected render() {
```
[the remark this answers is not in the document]
rkaraivanov: Agreed. Let us setup a meeting and discuss it. Having some coding guidelines will be a benefit.

src/components/rating/rating.ts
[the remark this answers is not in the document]
gmurray81: It's best if we avoid properties with union/intersection types on the public api, since I don't have a good way to represent them on the public apu other than using object at the c# level.

It would generally be better if properties with multiple types were multiple properties that interacted in a well defined way.

And for a callback style property, if it is formulated as an event, then it will automatically work for Blazor.

Do note though, that for Blazor server, the developer is unable to respond to events synchronously, so it would be best if we avoided putting critical functionality into an event that needs to be responded to synchronously.

So having something declaratice like emptySymbol/symbol would be much friendlier.

The other benefit of making things declarative is that they become much easier to specify from markup, so it's usually a win/win.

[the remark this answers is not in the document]
rkaraivanov: @gmurray81 
Does this mean that all component properties that accept a JS callback (e.g. custom rendering) should be transformed to event emitters?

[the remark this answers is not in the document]
rkaraivanov: @DiyanDimitrov 
After a discussion with the design team, we came to the agreement to not exposed an empty symbol property
as it will look 'broken' when the step property is different from 1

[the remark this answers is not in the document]
gmurray81: ideally when a callback is necessary its best to use an actual event, since then it will automatically be available at the blazor level, @rkaraivanov. If it's a scenario where it needs to be a callback instead, then we can still make it work, but its certainly easier if it's the former.

The more important thing to keep in mind is the with blazor we can't synchronously respond to an event or callback being fired and provide a response before the control returns to javascript. This is because in the Blazor Server model, the .net event handler code will actually be running on the server instead of in WebAssembly, so the logic may not even be on the client, so we are not able to make the js thread wait for the handler to process. So if you design something where the user _must_ provide a response in an event/callback , it will need a different way to operate in blazor.

We do have a fallback method where the user can actually configure a JS handler to be hooked up to an event/callback, instead of providing the handler in .NET, and that will be able to execute on the client. But since we try to avoid the developer needing to write JS for the Blazor product except if forced to, then we should generally try to avoid scenarios where this is required. This has been pretty tricky, because sometimes a callback/event is the easiest and simplest approach. We've had to jump through a lot of hoops to create declarative style approaches to avoid having to rely too much on events/callbacks. But as I said above, this effort isn't really wasted because it means that more power can be expressed declaratively, and maybe even from markup, rather than requiring the user to hook handlers up to the element.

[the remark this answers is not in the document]
DiyanDimitrov: @rkaraivanov,
IMO It would look 'broken' only if the icon and empty icon are completely different, which is a developer responsibility. This [sample](https://mui.com/components/rating/#customization) demonstrates rating component with precision, icon and empty icon and doesn't look broken.

src/components/rating/rating.ts
```diff
@@ -0,0 +1,297 @@
...
+  protected renderFractionWrapper(styles: { width: string }) {
+    return this.hoverPreview
+      ? html` <div
+          @click=${this.handleClick}
+          @mouseenter=${this.handleMouseEnter}
+          @mouseleave=${this.handleMouseLeave}
+          @mousemove=${this.handleMouseMove}
+        >
+          <div style=${styleMap(styles)} part="fraction ${this.size}">
+            <div part="symbols-wrapper fraction">${this.renderSymbols()}</div>
```
DiyanDimitrov, line 253: These two elements have `fraction` part. In the first element it is a main part name while in the second it is a modifier.

src/components/rating/rating.ts
```diff
@@ -0,0 +1,297 @@
...
+        `;
+  }
+
+  protected override render() {
+    const value = this.hoverState ? this.hoverValue : this.value;
+    const styles = { width: `${Math.round((value / this.max) * 100)}%` };
+
+    return html`
+      <div part="rating">
+        <label part="label ${this.size}" for="">${this.label}</label>
```
DiyanDimitrov, line 273: Do we need to provide built-in label? We don't provide such for the slider component.

src/components/rating/rating.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+
+  protected renderFractionWrapper(styles: { width: string }) {
+    return this.hoverPreview
+      ? html` <div
+          @click=${this.handleClick}
+          @mouseenter=${this.handleMouseEnter}
+          @mouseleave=${this.handleMouseLeave}
+          @mousemove=${this.handleMouseMove}
```
DiyanDimitrov, line 250: We could conditionally attach the handlers instead of duplicating the template.
```suggestion
          @mouseenter=${this.hoverPreview ? this.handleMouseEnter : null}
          @mouseleave=${this.hoverPreview ? this.handleMouseLeave : null}
          @mousemove=${this.hoverPreview ? this.handleMouseMove : null}
```

DiyanDimitrov (changes requested):
IMO the default value of the symbol gives the rating a Win98 look. I still think we should provide `symbol` and `emptySymbol` properties with default values of filled star (★) for symbol and outlined star (☆) for empty symbol.

gmurray81 (dismissed):

src/components/rating/rating.ts
```diff
@@ -0,0 +1,297 @@
...
+          </div>
+        `;
+  }
+
+  protected override render() {
+    const value = this.hoverState ? this.hoverValue : this.value;
+    const styles = { width: `${Math.round((value / this.max) * 100)}%` };
+
+    return html`
+      <div part="rating">
```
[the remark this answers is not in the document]
DiyanDimitrov: This comment is not addressed.

src/components/rating/rating.ts
DiyanDimitrov: This comment is not addressed.

src/components/rating/rating.ts
DiyanDimitrov: This comment is not addressed.

src/components/rating/rating-symbol.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @element igc-rating-symbol
+ *
+ * @slot - Default slot for projected symbols/icons.
+ */
+export default class IgcRatingSymbolComponent extends LitElement {
+  public static readonly tagName = 'igc-rating-symbol';
+
+  protected override render() {
+    return html`<slot></slot>`;
+  }
```
DiyanDimitrov, line 16: Shouldn't we use light DOM instead?

src/components/rating/rating.ts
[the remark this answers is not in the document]
rkaraivanov: @desig9stein 
Can you chime in on this?

src/components/rating/rating.ts
rkaraivanov: @desig9stein 
Can you chime in on this?

desig9stein: @DiyanDimitrov I am not sure if I understand you correctly. What is the problem being the way it is? Why do you call it modified, we don't use BEM methodology in the web components. However I optimize the styles a bit, please take a look now.

src/components/rating/rating.ts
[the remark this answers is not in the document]
desig9stein: @DiyanDimitrov 10x for the suggestion, please take a look at it now.

src/components/rating/rating.ts
desig9stein: Maybe @StefanIvanov can answer that question.

StefanIvanov: If you look at the [rating spec](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Rating-Specification) you will notice that we have in the designs a label and even a value label, but then we decided that both might be a bit too much and left the only the more agnostic one i.e. the label. I understand your concerns and reference to the slider as it practically has the opposite i.e. no such label but one or two value labels positioned on the thumbs. Now the slider design is largely affected by the default material theme where labels of form elements are subtle and to a degree unnecessary. Having added bootstrap and fluent themes they are much more prominent when it comes to labels of form elements and this is also a mechanism to establish a rhythm of the form itself. Therefore, I'd advise considering the addition of a "generic" label to the slider instead of removing it from the rating.

src/components/rating/rating.ts
DiyanDimitrov: My idea was to not have two elements with the same part name. This is only acceptable if a part name is a state modifier such as `selected`, `active`, etc. I am ok with the latest implementation.

DiyanDimitrov (approved):

DiyanDimitrov (approved):

src/components/rating/rating.ts
```diff
@@ -0,0 +1,322 @@
...
+/**
+ * Rating provides insight regarding others' opinions and experiences,
+ * and can allow the user to submit a rating of their own
+ *
+ * @element igc-rating
+ *
+ * @fires igcChange - Emitted when the value of the control changes.
+ * @fires igcHover - Emitted when hover is enabled and the user mouses over a symbol of the rating.
+ *
+ * @csspart base - The main wrapper which holds all of the rating elements.
```
DiyanDimitrov, line 36: We should add the other exposed parts here.

desig9stein: @DiyanDimitrov I've added the rest of the parts.

PR #140, opened 2021-11-18
fix: Setting avatar src correctly resets error state

### LINKED ISSUE #138: Avatar's hasError state is not reset when src is changed
## Description  
Avatar's hasError state is not reset when src is changed.

## Steps to reproduce  

1. Open the avatar story.
2. Enter invalid url for src.
3. Enter valid url for src.

## Result  
The valid url is not loaded.

## Expected result  
The valid url should be loaded.

rkaraivanov:
Closes #138

### DIFF
--- a/src/components/avatar/avatar.ts
+++ b/src/components/avatar/avatar.ts
@@ -25,9 +25,18 @@ export default class IgcAvatarComponent extends SizableMixin(LitElement) {
   /** @private */
   public static styles = [styles];
 
+  public set src(val: string) {
+    const oldVal = this._src;
+    this._src = val;
+    this.hasError = false;
+    this.requestUpdate('src', oldVal);
+  }
+
   /** The image source to use. */
   @property()
-  public src!: string;
+  public get src() {
+    return this._src;
+  }
 
   @state()
   private hasError = false;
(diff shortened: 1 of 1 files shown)


src/components/avatar/avatar.ts
```diff
@@ -25,9 +25,18 @@ export default class IgcAvatarComponent extends SizableMixin(LitElement) {
   /** @private */
   public static styles = [styles];
 
+  public set src(val: string) {
```
DiyanDimitrov, line 28: What do you think about using the `watch` decorator instead custom get/set?

rkaraivanov: Yes, but then I would have to create a separate method for resetting the error state and decorate it.
This method will not be called anywhere so I would have to make the linter to ignore it, as it is an error otherwise.
I think this a bit clearer in this scenario.

DiyanDimitrov: If the method is protected the linter would be fine with it. Personally I think a custom get/set looks more complicated and is error-prone since you may forgot to call `requestUpdate`.

DiyanDimitrov (approved):

PR #158, opened 2021-11-29
feat(checkbox): implement-dark-themes

SisIvanova:


simeonoff (commented):
We shouldn't copy the entire source code for each light theme. Rather include the light mixin and modify just the parts that need modifying.

src/components/checkbox/checkbox.dark.fluent.scss
```diff
@@ -1,34 +1,20 @@
 @use '../../styles/utilities' as utils;
+@use 'checkbox.fluent.scss' as light-checkbox;
+
+@include light-checkbox.theme();
```
simeonoff, line 4: This should rather be included in the theme mixin.

simeonoff (changes requested):

simeonoff (approved):

simeonoff (approved):

simeonoff (approved):

PR #167, opened 2021-12-01
feat: (scrollbars) add theme pre component

desig9stein:


rkaraivanov:
@simeonoff @desig9stein 
What is the state of this item?

simeonoff:
It looks like it's become stale. We'll discuss today and let you know.

PR #202, opened 2022-01-05
fix: Checkbox/switch validity status

### LINKED ISSUE #200: Checkbox selection focuses its input
## Description  
When checkbox state is changed to selected its input is focused.

## Expected result  
Selection should not change the focused element.

## Scenario
1. Checkbox is part of the igc-tree-item.
2. An igc-tree-item is focused.
3. Clicking space should change its selection state but the focused element should remain the same.

## Attachments  
image

### LINKED ISSUE #201: setting required on checkbox in form does not ensure it is checked
## Description  
When adding a checkbox with required set in a form, when submitting the form no error is throw if the checkbox is not checked

## Expected result  
If the checkbox is listed as required it should need to be checked in order for the submit to complete successfully, for example in the form samples (https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/form) adding required should be the way that the form ensures that the user has "accepted the license agreement".

rkaraivanov:
Closes #200 
Closes #201


src/components/checkbox/switch.ts
```diff
@@ -36,38 +36,37 @@ export default class IgcSwitchComponent extends IgcCheckboxBaseComponent {
 
   private handleClick() {
     this.checked = !this.checked;
+    this.emitEvent('igcChange', { detail: this.checked });
```
gedinakova, line 39: By moving the event emitter here, we will stop emitting when changing the value via keyboard.

src/components/checkbox/checkbox.ts
```diff
@@ -39,57 +39,55 @@ export default class IgcCheckboxComponent extends IgcCheckboxBaseComponent {
   @property({ type: Boolean, reflect: true })
   public indeterminate = false;
 
-  private handleClick() {
+  protected handleClick() {
     this.checked = !this.checked;
     this.indeterminate = false;
+    this.emitEvent('igcChange', { detail: this.checked });
```
gedinakova, line 45: By moving the event emitter here, we will stop emitting when changing the value via keyboard.

src/components/checkbox/checkbox.spec.ts
```diff
@@ -103,6 +103,23 @@ describe('Checkbox', () => {
...
+        html`<igc-checkbox required checked></igc-checkbox>`
+      );
+      expect(el.reportValidity()).to.equals(true);
+
+      el = await fixture<IgcCheckboxComponent>(
+        html`<igc-checkbox></igc-checkbox>`
+      );
+      expect(el.reportValidity()).to.equals(true);
+    });
+
```
gedinakova, line 122: We should add a test for igcChange on keyboard interaction (currently we test only on click - last test).

src/components/checkbox/switch.spec.ts
```diff
@@ -101,6 +101,21 @@ describe('Switch', () => {
...
+
+      el = await fixture<IgcSwitchComponent>(
+        html`<igc-switch required checked></igc-switch>`
+      );
+      expect(el.reportValidity()).to.equals(true);
+
+      el = await fixture<IgcSwitchComponent>(html`<igc-switch></igc-switch>`);
+      expect(el.reportValidity()).to.equals(true);
+    });
+
```
gedinakova, line 118: We should add a test for igcChange on keyboard interaction (currently we test only on click - last test).

gedinakova (changes requested):

src/components/checkbox/checkbox.spec.ts
rkaraivanov: Since browsers fire `click` after `keyup` (which should be called _activate_, imo) on Enter/Spacebar when the focus is on the input, this would be hard to emulate in the test environment with firing events programmatically.
We can assume the click handler tests cover this scenario until browsers drop the quirksmode (doubtful about that).

gedinakova (approved):

simeonoff (approved):

PR #212, opened 2022-01-18
feat(radio): support required attribute

### LINKED ISSUE #203: IgcRadio component should support 'required' attribute
As per [MDN spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) the radio input should support required.

onlyexeption:
Closes #203


rkaraivanov (approved):
While testing the control in an _igc-form_ I found several problems with how the control behaves with validation and navigation.
I will log them in a separate issue as to not bog down this one.

src/components/radio/radio.spec.ts
```diff
@@ -158,6 +160,19 @@ describe('Radio Component', () => {
       );
     });
 
+    it('sets the required property successfully', async () => {
+      radio.required = true;
+      expect(radio.required).to.be.true;
+      await elementUpdated(radio);
+      expect(input).dom.to.equal(`<input required />`, DIFF_OPTIONS);
```
gedinakova, line 167: Maybe we can check whether the asterisk is shown in the label?

gedinakova (changes requested):

gedinakova (changes requested):
Add the change to CHANGELOG.md

gedinakova (approved):

src/components/radio/radio.ts
```diff
@@ -128,9 +132,16 @@ export default class IgcRadioComponent extends EventEmitterMixin<
   @watch('checked', { waitUntilFirstUpdate: true })
   protected handleChange() {
     if (this.checked) {
+      if (this.getSiblings().some((r) => r.required)) {
```
gedinakova, line 135: Isn't the setRequired() in the radio group enough? Do we need this additional code here?

gedinakova (changes requested):

rkaraivanov (changes requested):
```typescript
  @queryAssignedNodes(undefined, true, 'igc-radio')
  private _slottedRadios!: NodeListOf<IgcRadioComponent>;


  private get radios() {
    return Array.from(this._slottedRadios).filter((radio) => !radio.disabled);
  }
```
Since Lit 2.1.x you can remove the _radios_ getter and simplify this to
```typescript
@queryAssignedElements({
    flatten: true,
    selector: 'igc-radio:not([disabled])',
  })
  private radios!: Array<IgcRadioComponent>;
```

src/components/radio/radio.ts
```diff
@@ -128,9 +132,16 @@ export default class IgcRadioComponent extends EventEmitterMixin<
...
+      }
+
       this.getSiblings().forEach((radio) => {
         radio.checked = false;
         radio._tabIndex = -1;
+        if (this.required) {
+          radio.required = false;
+        }
       });
       this.input.focus();
```
rkaraivanov, line 146: Do we really need this here? Since by default the tabindex is 0 and we update the rest of the siblings when a selection in the group is made this seems like a no-op

onlyexeption: Removing this line seems to cause a problem with setting the current radio as the activeElement

gedinakova (dismissed):

gedinakova:
@onlyexeption @rkaraivanov Currently, after resetting the form, the required attribute remains on the last selected item which causes the popup to show on this item instead of the first one.

gedinakova:
> @onlyexeption @rkaraivanov @simeonoff  Currently, after resetting the form, the required attribute remains on the last selected item which causes the popup to show on this item instead of the first one.

This seems to be an issue with the radio's igcChange event being emitted only when checked is true.
@watch('checked', { waitUntilFirstUpdate: true })
  protected handleChange() {
    if (this.checked) {
      this.getSiblings().forEach((radio) => {
        radio.checked = false;
        radio._tabIndex = -1;
      });
      this.input.focus();
      this._tabIndex = 0;
      this.emitEvent('igcChange', { detail: this.checked });
    }
  }

I suggest we move the emitEvent outside because this is not a igcChecked event, so the change can be to false as well.

simeonoff:
> > @onlyexeption @rkaraivanov @simeonoff  Currently, after resetting the form, the required attribute remains on the last selected item which causes the popup to show on this item instead of the first one.
> 
> This seems to be an issue with the radio's igcChange event being emitted only when checked is true.
> 
> ```ts
> @watch('checked', { waitUntilFirstUpdate: true })
>   protected handleChange() {
>     if (this.checked) {
>       this.getSiblings().forEach((radio) => {
>         radio.checked = false;
>         radio._tabIndex = -1;
>       });
>       this.input.focus();
>       this._tabIndex = 0;
>       this.emitEvent('igcChange', { detail: this.checked });
>     }
>   }
> ```
> 
> I suggest we move the emitEvent outside because this is not an igcChecked event, so the change can be to false as well.

This means the event will be fired twice.

gedinakova:
> > > @onlyexeption @rkaraivanov @simeonoff  Currently, after resetting the form, the required attribute remains on the last selected item which causes the popup to show on this item instead of the first one.
> > 
> > 
> > This seems to be an issue with the radio's igcChange event being emitted only when checked is true.
> > ```ts
> > @watch('checked', { waitUntilFirstUpdate: true })
> >   protected handleChange() {
> >     if (this.checked) {
> >       this.getSiblings().forEach((radio) => {
> >         radio.checked = false;
> >         radio._tabIndex = -1;
> >       });
> >       this.input.focus();
> >       this._tabIndex = 0;
> >       this.emitEvent('igcChange', { detail: this.checked });
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suggest we move the emitEvent outside because this is not an igcChecked event, so the change can be to false as well.
> 
> This means the event will be fired twice.

We took a different approach!

gedinakova (dismissed):

gedinakova (approved):

rkaraivanov (approved):

PR #214, opened 2022-01-19
Changed boolean properties to default to false.

### LINKED ISSUE #199: Replace the boolean attributes with default value `true`
Currently we have some components that have boolean attributes with default value `true` like `IgcCardComponent.outlined`. It is not possible to remove these attributes in the `html`. We have to replace them with attributes which has default value `false` for example `IgcCardComponent.elevated`.

gedinakova:
Closes #199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Fix checkbox/switch validity status
-- Renamed Card component's `outlined` property to `elevated`.
-- Renamed Calendar component's `hasHeader` property & `has-header` attribute to `hideHeader` & `hide-header` respectively.
+- Replaced Card component's `outlined` property with `elevated`.
+- Replaced Calendar component's `hasHeader` property & `has-header` attribute with `hideHeader` & `hide-header` respectively.
 
 ## 1.0.0 - 2021-11-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Fix checkbox/switch validity status
+- Renamed Card component's `outlined` property to `elevated`.
+- Renamed Calendar component's `hasHeader` property & `has-header` attribute to `hideHeader` & `hide-header` respectively.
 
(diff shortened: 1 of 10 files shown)


src/components/card/card.material.scss
```diff
@@ -4,26 +4,21 @@
...
+:host([elevated]) {
+    border: none;
+    box-shadow: var(--igc-elevation-4);
+    backface-visibility: hidden;
+    transition: box-shadow .1s ease-out;
 
     &:hover,
     &:focus-within {
-        box-shadow: none;
+        box-shadow: var(--igc-elevation-6);
```
DiyanDimitrov, line 22: The hover styling when elevated is set and the theme is material is not working. I was able to make it work like this:
```scss
:host([elevated]) {
    border: none;
    box-shadow: var(--igc-elevation-4);
    backface-visibility: hidden;
    transition: box-shadow .1s ease-out;
}

:host([elevated]:hover),
:host([elevated]:focus-within) {
    box-shadow: var(--igc-elevation-6);
}
```

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Fix checkbox/switch validity status
+- Renamed Card component's `outlined` property to `elevated`.
+- Renamed Calendar component's `hasHeader` property & `has-header` attribute to `hideHeader` & `hide-header` respectively.
```
DiyanDimitrov, line 20: Since the properties are not just renamed but their logic is inverted I would suggest:
```suggestion
- Replaced Card component's `outlined` property with `elevated`.
- Replaced Calendar component's `hasHeader` property & `has-header` attribute with `hideHeader` & `hide-header` respectively.
```

DiyanDimitrov (changes requested):

CHANGELOG.md
DiyanDimitrov: Also maybe we need to highlight it as a breaking change.

DiyanDimitrov: Thinking about this further we may want to first deprecate the old props as [Semantic Versioning](https://semver.org/spec/v2.0.0.html#how-should-i-handle-deprecating-functionality) suggests.

DiyanDimitrov (approved):

desig9stein (approved):

PR #217, opened 2022-01-20
fix(nav-drawer): remove events

### LINKED ISSUE #210: `-ing` events should be cancelable instead of using `cancel` event argument
`-ing` events should be canceled using the preventDefault() instead of exposing a `cancel` event argument.

onlyexeption:
Closes #210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Fix checkbox/switch validity status
+- Use preventDefault() to cancel Navigation Drawer events instead of using a `cancel` event argument.
 
 ## 1.0.0 - 2021-11-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Removed
 
-- Events in Navigation drawer component.
+- Removed `igcOpening`, `igcOpened`, `igcClosing` and `igcClosed` events from Navigation drawer component.
 
 ## 1.0.0 - 2021-11-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Fix checkbox/switch validity status
-- Use preventDefault() to cancel Navigation Drawer events instead of using a `cancel` event argument.
+- Remove events in Navigation Drawer components.
+
+### Removed
+- Events in Navigation drawer component.
 
 ## 1.0.0 - 2021-11-22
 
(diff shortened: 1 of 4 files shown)


DiyanDimitrov (approved):

DiyanDimitrov (changes requested):
Update changelog.

src/components/nav-drawer/nav-drawer.spec.ts
```diff
@@ -165,7 +165,7 @@ describe('Navigation Drawer', () => {
 
     it('should successfully cancel opening event', async () => {
       el.addEventListener('igcOpening', (ev: CustomEvent) => {
-        ev.detail.cancel = true;
+        ev.preventDefault();
```
gedinakova, line 168: The test has 2 checks at the end:
      expect(eventSpy).calledWith('igcOpening');
      expect(el.open).to.equal(false);
The first one would pass even if you don't prevent the event. I suggest we change it to:
expect(eventSpy).calledOnceWith('igcOpening');
This way the check would fail because emitEvent would be called twice - once with igcOpening & once with igcOpened.
The other option of course is to just remove this spy check.

src/components/nav-drawer/nav-drawer.spec.ts
```diff
@@ -181,7 +181,7 @@ describe('Navigation Drawer', () => {
     it('should successfully cancel closing events', async () => {
       el.show();
       el.addEventListener('igcClosing', (ev: CustomEvent) => {
-        ev.detail.cancel = true;
+        ev.preventDefault();
       });
 
       const eventSpy = sinon.spy(el, 'emitEvent');
```
gedinakova, line 187: Same as above.

gedinakova (changes requested):

gedinakova (approved):

gedinakova:
We will need to discuss whether event are needed at all.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Replaced Calendar component's `hasHeader` property & `has-header` attribute with `hideHeader` & `hide-header` respectively.
 - Replaced Card component's `outlined` property with `elevated`.
 
+### Removed
+
+- Events in Navigation drawer component.
```
gedinakova, line 25:
```suggestion
- Removed `igcOpening`, `igcOpened`, `igcClosing` and `igcClosed` events from Navigation drawer component.
```

gedinakova (changes requested):

DiyanDimitrov (changes requested):
There are some build errors:
```
Error: src/components/nav-drawer/nav-drawer.spec.ts(8,1): error TS6133: 'sinon' is declared but its value is never read.
Error: src/components/nav-drawer/nav-drawer.ts(3,1): error TS6133: 'Constructor' is declared but its value is never read.
Error: src/components/nav-drawer/nav-drawer.ts(4,1): error TS6133: 'EventEmitterMixin' is declared but its value is never read.
```

gedinakova (approved):

DiyanDimitrov (approved):

PR #218, opened 2022-01-21
fix(slider): remove z-index

DiyanDimitrov:
The slider track was not appearing in the samples repo. In order to fix it I have removed the z-index an made sure the elements are in the correct order in the DOM.

### DIFF
--- a/src/components/slider/slider.material.scss
+++ b/src/components/slider/slider.material.scss
@@ -45,7 +45,6 @@ $thumb-label-height: rem(30px);
     background: color(secondary, 500, .24);
     transition: background .2s ease-out;
     border-radius: inherit;
-    z-index: -1;
 }
 
 [part='steps'] {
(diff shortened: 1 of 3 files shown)


src/components/slider/slider.material.scss
```diff
@@ -187,7 +183,6 @@ $thumb-label-height: rem(30px);
     color: contrast-color(gray, 700);
     background: color(gray, 700);
     opacity: 0;
-    z-index: 1;
```
SisIvanova, line 190: If you remove that one, the thumb ripple on hover and focus will be displayed above the thumb-label.
Screen Shot 2022-01-24 at 11 10 44 AM
Screen Shot 2022-01-24 at 11 09 20 AM

DiyanDimitrov: I have swapped the elements in the template so that they appear correctly.

SisIvanova (approved):

PR #226, opened 2022-01-31
feat(snackbar): implement snackbar component

SisIvanova:


StefanIvanov:
The indigo theme has poor contrast ration of the button gray 700 and the background of the snackbar using the same color with 62% opacity. Can we make the button label use a contrast function or at least be white as that would be the output of it anyhow?

StefanIvanov:
I spoke with Simeon and after confirming with the app builder design system, it is best to change the background color to surface and leave the text and button use gray 700

src/components/snackbar/themes/snackbar.indigo.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../../styles/utilities' as *;
+
+[part~='base'] {
+    @include border-radius(rem(6px));
+
+    background: color(gray, 700, .62);
+    color: contrast-color(gray, 700);
```
StefanIvanov, line 7:
```suggestion
    background: color(surface, 500);
    color: color(gray, 700);
```

src/components/snackbar/snackbar.ts
```diff
@@ -0,0 +1,117 @@
...
+          ? html`
+              <igc-button
+                variant="flat"
+                part="action"
+                @click=${this.handleClick}
+              >
+                ${ifDefined(this.actionText)}
+              </igc-button>
+            `
+          : ''}
```
simeonoff, line 107: Use the [`nothing`](https://lit.dev/docs/api/templates/#nothing) value here instead of `''`;
```suggestion
          : nothing}
```

src/components/snackbar/themes/snackbar.bootstrap.scss
```diff
@@ -0,0 +1,10 @@
+@use '../../../styles/utilities' as *;
+
+[part~='base'] {
+    background: color(surface, 500);
+    color: contrast-color(surface, 500);
+    border: 1px solid color(gray, 300);
+    width: rem(348px);
```
simeonoff, line 7: I am not sure about this. Maybe it should be max-width, instead of width?

src/components/snackbar/themes/snackbar.fluent.scss
```diff
@@ -0,0 +1,9 @@
+@use '../../../styles/utilities' as *;
+
+[part~='base'] {
+    background: color(gray, 100);
+    color: contrast-color(gray, 100);
+    border-radius: 0;
+    width: rem(352px);
```
simeonoff, line 7: Same comment about width here.

src/components/snackbar/themes/snackbar.indigo.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../../styles/utilities' as *;
+
+[part~='base'] {
+    @include border-radius(rem(6px));
+
+    background: color(surface, 500);
+    color: color(gray, 700);
+    padding: rem(10px) rem(16px);
+    width: rem(352px);
```
simeonoff, line 9: And here.

src/components/snackbar/snackbar.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** Determines whether the snackbar is opened. */
+  @property({ type: Boolean, reflect: true })
+  public open = false;
+
+  /** Determines the duration in ms in which the snackbar will be visible. */
+  @property({ type: Number, attribute: 'display-time' })
+  public displayTime!: number;
+
+  /** Determines whether the snackbar should close after the displayTime is over. */
+  @property({ type: Boolean, reflect: true, attribute: 'keep-open' })
```
simeonoff, line 53: Does this need to reflect? Also types are inferred, no need to declare them explicitly.

src/components/snackbar/themes/snackbar.material.scss
```diff
@@ -0,0 +1 @@
+@use '../../../styles/utilities' as *;
```
simeonoff, line 1: This file should be scrapped altogether. No need to have an empty file just for the sake of having a material theme.

simeonoff (changes requested):

src/components/snackbar/snackbar.ts
simeonoff: Why did you mark it as resolved? It's clearly not been changed!?

simeonoff (changes requested):
See comments.

src/components/snackbar/themes/snackbar.bootstrap.scss
didimmova: The width should be fixed - see [material snackbar](https://material.io/components/snackbars#specs)

simeonoff: I don't think so. I see a bunch of examples where the UI is much larger than the set 348 pixels. @StefanIvanov please reconsider.

simeonoff (dismissed):

simeonoff (approved):

PR #232, opened 2022-02-03
Refactor Component Themes

### LINKED ISSUE #228: Allow Composable Component
## Description

So far all themes, except the material one, which is the one all components encapsulate, are included via a global theme. This works well for simple components but we face an issue when more complex components use other components as part of their template. In those instances, the global themes cannot penetrate the shadow DOM and thus, internal components remain with the default material theme.

## Proposal

Components should dynamically switch the encapsulated styles based on some global configuration. This could happen via a `theme` property on each component or in another more generic way.

## TODO(s):

- [ ] Implement a mechanism to allow for dynamic theme switching on the component level.

Refactor affected component themes so all variants use the shadow DOM selectors for styling. Those components are:
- [x] Badge - @didimmova 
- [x] Button - @didimmova 
- [x] Calendar - @desig9stein 
- [x] Card - @SisIvanova 
- [x] Checkbox - @SisIvanova  
- [x] Icon Button - @didimmova 
- [x] Input - @simeonoff 
- [x] List - @didimmova 
- [x] Navbar - @SisIvanova 
- [x] Nav Drawer - @desig9stein 
- [x] Radio Button - @didimmova
- [x] (Range) Slider - @SisIvanova 
- [x] Switch - @SisIvanova

simeonoff:
Make all component themes work in the shadow DOM.
Closes #228

Please see the [Styles & Themes Wiki page](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Styling-and-Themes) for more information as to how the implementation in this PR works.


rkaraivanov (approved):

rkaraivanov (approved):

src/theming/config.ts
```diff
@@ -0,0 +1,55 @@
+import { ChangeThemeEventDetail, CHANGE_THEME_EVENT } from './theming-event';
+import { Theme } from './types';
+import { getAllCSSVariables } from './utils';
+
+let theme: Theme;
+
+type ThemeConfig = {
+  theme: Theme;
+  variant?: 'light' | 'dark';
```
DiyanDimitrov, line 9: Do we need `variant`? It seems that we are not using it anywhere.

rkaraivanov: AFAIK, we talked with @simeonoff that if we have to extend the API in the future and to pass the variant as well as the theme, it is better to have a single argument with optional props than to change the signature of the `configureTheme` function.

rkaraivanov (approved):

src/theming/config.ts
DiyanDimitrov: We could add `variant` as optional parameter to the function which won't be a breaking change. I don't think we will have any other options for this function.

simeonoff: True, I guess it's just a matter of _named_(config with typed signature) arguments vs _positional_(function args) ones. Whatever you say, I can go back to simply passing a string literal to the `configureTheme` function.

DiyanDimitrov: Yep, it depends on the personal preferences. For up to 2-3 parameters I prefer the `func(arg1, arg2)` approach. 

Since we are going to release with a single parameter I would prefer `configureTheme('material')` than `configureTheme({ theme: 'material' })`.

DiyanDimitrov: Also if we decide that we need a more complex config in the future we can change the signature like this:

```ts
configureTheme(theme: string | ThemeConfig)
```

rkaraivanov: > Also if we decide that we need a more complex config in the future we can change the signature like this:
> 
> ```ts
> configureTheme(theme: string | ThemeConfig)
> ```

I still think is vastly better to be 
```typescript 
configureTheme(options: ThemeConfig)
```
and add new props to the options object than to union the type at a later stage. 
But as you pointed out it is a matter of personal preference.

simeonoff: @rkaraivanov and @DiyanDimitrov, reverted it to the original implementation. The `configureTheme` expects an argument of type Theme now.  Obviously, if in JavaScript this will be a simple string.

rkaraivanov (approved):

Issue #55, opened 2021-08-30
DropDown component


gedinakova:
_Estimated Delivery Sprint - Sprint 4_

- [x] Research
- [x] Specification
- [x] Component and tests implementation
- [x] Design implementation
- [x] API Documentation

PR #238, opened 2022-02-07
feat(Dropdown): Adding component

gedinakova:
#55


stories/dropdown.stories.ts
```diff
@@ -0,0 +1,294 @@
...
+export default metadata;
+interface ArgTypes {
+  placement: IgcPlacement;
+  open: boolean;
+  strategy: 'absolute' | 'fixed';
+  flip: boolean;
+  closeOnOutsideClick: boolean;
+  scrollStrategy: 'scroll' | 'close' | 'block' | 'noop';
+}
+// end region
```
DiyanDimitrov, line 80: It should be `endregion`.

stories/dropdown.stories.ts
```diff
@@ -0,0 +1,294 @@
...
+  'Specification',
+  'Implementation',
+  'Testing',
+  'Samples',
+  'Documentation',
+  'Builds',
+];
+const Template: Story<ArgTypes, Context> = (
+  {
+    flip = false,
```
DiyanDimitrov, line 120: Why don't we bind to the `open` property?

src/components/dropdown/dropdown.ts
```diff
@@ -0,0 +1,579 @@
...
+  /** Sets the open state of the dropdown list. */
+  @property({ type: Boolean })
+  public open = false;
+
+  /** Sets the dropdown list's positioning strategy. */
+  @property({ attribute: false })
+  public positionStrategy: 'absolute' | 'fixed' = 'absolute';
+
+  /** The preferred placement of the dropdown list around the target element. */
+  @property({ attribute: false })
```
DiyanDimitrov, line 85: Since we are using union type we should provide the options in the API comment like this `@type {"top" | "bottom" | ...}`

DiyanDimitrov (changes requested):

src/components/dropdown/dropdown.ts
```diff
@@ -0,0 +1,577 @@
...
+      placement: this.placement,
+      positionStrategy: this.positionStrategy,
+      flip: this.flip,
+      closeOnOutsideClick: this.closeOnOutsideClick,
+      offset: this.offset,
+    };
+
+    this.toggleController = new IgcToggleController(
+      this,
+      this.target ?? this.target,
```
DiyanDimitrov, line 168: I guess it should be just `this.target`.

src/components/dropdown/dropdown.ts
```diff
@@ -0,0 +1,577 @@
...
+
+  @queryAssignedElements({ slot: 'target' })
+  private targetNodes!: Array<HTMLElement>;
+
+  /** Sets the open state of the dropdown list. */
+  @property({ type: Boolean })
+  public open = false;
+
+  /** Sets the dropdown list's positioning strategy.
+   * @type {"absolute" | "fixed"}
```
DiyanDimitrov, line 81: In this case defining `@type` is not necessary.

src/components/dropdown/dropdown.ts
```diff
@@ -0,0 +1,577 @@
...
+   */
+  @property({ attribute: false })
```
DiyanDimitrov, line 83: Why do we set attribute to false for almost all dropdown properties?

src/components/dropdown/dropdown.ts
```diff
@@ -96,24 +100,28 @@ export default class IgcDropDownComponent extends EventEmitterMixin<
   @property({ type: Boolean })
   public closeOnOutsideClick = true;
 
-  /** Determines the behavior of the dropdown list during scrolling the container. */
+  /** Determines the behavior of the dropdown list during scrolling the container.
+   *  @type {"scroll" | "close" | "block" | "noop"}
```
DiyanDimitrov, line 104: Defining `@type` is not necessary here too.

stories/dropdown.stories.ts
gedinakova: There's a problem with the placement property. The autogeneration removes it every time, so to prevent auto-generation I intentionally mistyped the 'endregion'.

src/components/dropdown/dropdown.ts
```diff
@@ -0,0 +1,569 @@
...
+  @property({ type: Boolean })
+  public closeOnOutsideClick = true;
+
+  /** Determines the behavior of the dropdown list during scrolling the container. */
+  @property()
+  public scrollStrategy: 'scroll' | 'close' | 'block' | 'noop' = 'noop';
+
+  /** The amount of offset in horizontal and/or vertical direction. */
+  @property()
+  public offset!: { x: number; y: number };
```
DiyanDimitrov, line 107: We could possibly implement custom attribute converter for this one that uses comma-separated string.

DiyanDimitrov (approved):

PR #249, opened 2022-02-22
feat: Mask input 

### LINKED ISSUE #173: Masked input component
_Estimated Delivery FEB 22

- [x] Research
- [x] Specification
- [ ] API Documentation
- [ ] Design implementation

rkaraivanov:
Closes #173


simeonoff (dismissed):

ChronosSF (dismissed):

src/components/mask-input/mask-input.spec.ts
```diff
@@ -0,0 +1,498 @@
...
+    it('empty value without literals', async () => {
+      expect(masked.value).to.be.undefined;
+    });
+
+    it('empty value with literals', async () => {
+      masked.withLiterals = true;
+      syncParser();
+      await elementUpdated(masked);
+
+      expect(masked.value).to.equal(parser.apply());
```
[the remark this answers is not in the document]
damyanpetev: For the validation bit, mostly thinking about the case where the input is empty, required and still with literals and if that can mess the required check. From what I can see that won't happen since it only affects the value getter. Still might simply look odd getting a mask back on an empty input - it's still _value_ with literals, but there's no value.

As for the value itself, it currently behaves like "with literals and prompts". Technically, we can expose more options, but I'm torn on this - can't think of too many scenarios where the literals are wanted without the full mask filled (which should be achievable through validation), and even on partially filled at least you get the same fixed length value which should be simple enough to handle. Honestly, taking opinions on this one, could do either way.

src/components/mask-input/mask-input.ts
```diff
@@ -0,0 +1,360 @@
...
+
+  protected override handleBlur() {
+    this.hasFocus = false;
+    this.updateMaskedValue();
+    super.handleBlur();
+  }
+
+  protected override handleChange() {
+    this.emitEvent('igcChange', { detail: this.value });
+    this.invalid = !this.parser.isValidString(this._value);
```
damyanpetev, line 256: Since `parser.isValidString` will always return false for empty strings when there's a required position, it should only be checked if there is a value

src/components/mask-input/mask-input.ts
```diff
@@ -0,0 +1,360 @@
...
+    direction?: 'backward' | 'forward' | 'none'
+  ): void {
+    super.setSelectionRange(start, end, direction);
+    this.selection = { start, end };
+  }
+
+  /** Checks for validity of the control and shows the browser message if it's invalid. */
+  public reportValidity() {
+    return (
+      this.input.reportValidity() && this.parser.isValidString(this._value)
```
damyanpetev, line 308: Same here - only check required positions if there is something filled. Empty value is the responsibility of the `required` attribute rule and will be handled by the native `report/checkValidity`

src/components/mask-input/mask-input.ts
```diff
@@ -0,0 +1,360 @@
...
+      this.maskedValue = this.parser.apply(this._value);
+    }
+  }
+
+  @watch('required', { waitUntilFirstUpdate: true })
+  @watch('disabled', { waitUntilFirstUpdate: true })
+  @watch('value', { waitUntilFirstUpdate: true })
+  protected handleInvalidState() {
+    this.updateComplete.then(
+      () => (this.invalid = !this.input.checkValidity())
```
damyanpetev, line 126: This has the potential to override any previous invalid state from required mask positions if `required` is turned off for example as it will only check the native validity. I think you really need a single equivalent of `checkValidity` similar to the `reportValidity` for this to be consistent.

On second thought - both the existence of value and disabled should be taken into account when performing our checks for positions

damyanpetev (commented):
- The spec mentions `set an autofocus attribute so that the field can become focused on initial page load` but I don't see that being passed down to the native input, so it doesn't work
- When an empty input is `readonly` focusing it still brings up the mask with prompts, which I think shouldn't happen as it suggest it can be edited
- With the default mask if you fill some end portion (i.e. skip a few start positions) when you focus back the input the value shifts to the start. Bit edge, but still landed on it
- The input event is fired when deleting over literals that in a mask does nothing, but I'll chuck that under a minor gripe and can be left as is

src/components/mask-input/mask-input.ts
```diff
@@ -0,0 +1,380 @@
...
+      return this.input.checkValidity();
+    }
+
+    if (!this._value && this.required) {
+      return false;
+    }
+
+    return (
+      this.input.checkValidity() && this.parser.isValidString(this.input.value)
+    );
```
damyanpetev, line 338: Close but no cigar 😄
An empty input (`!this._value === true`) that is not required will still hit the parser validation and become invalid, when it shouldn't. Can be adjusted this way:
```suggestion
    if (!this._value) {
      return !this.required;
    }

    return (
      this.input.checkValidity() && this.parser.isValidString(this.input.value)
    );
```
This can be adjusted with a focus check to go through the native check if you'd like.

Honestly, no idea how much we should care about calling the native check validity when the only thing it can do is required and only when the mask is not focused. We don't explicitly support the invalid event it produces and before we implement ElementInternals it can't be made consistent anyway.

damyanpetev (approved):

simeonoff (approved):

ChronosSF (approved):

PR #253, opened 2022-02-24
feat(chip): create chip component

didimmova:


StefanIvanov:
I am leaving my comments:
Bootstrap theme:
Chip with default type when hovered should use grays 400 as background-color
Chip outline for focus should use the 200 variant of the color characterizing the type

Fluent theme:
Chip with default type should have grays 700 as idle color and on hover use the contrast-color with grays 200
Primary and Secondary on the light theme should be #2b579a as per theme definitions
Primary and Secondary on the dark theme should be #217346 as per theme definitions

StefanIvanov (approved):

PR #262, opened 2022-03-02
refactor(switch): update component to match latest boostrap spec

### LINKED ISSUE #254: Switch design in Bootstrap
## Description  
The Switch design is not accurate in the Bootstrap theme light and dark schema.

## Expected result  
The Switch should use the design from [Bootstrap docs](https://getbootstrap.com/docs/5.1/forms/checks-radios/#switches).

SisIvanova:
Closes #254


simeonoff (commented):
Looks good, however, can we make the thumb 10px instead of 11px, so that it doesn't look vertically off center due to the way sub-pixel rendering works on some Operating Systems.

PR #263, opened 2022-03-02
refactor(list): fix refactor on list themes

didimmova:


src/components/list/themes/light/list-header.indigo.scss
```diff
@@ -0,0 +1,8 @@
+@use './list-header.base.scss';
```
simeonoff, line 1: This should rather be imported and used as the default static styles on the list-header.ts. It shouldn't be imported here.

src/components/list/themes/light/list-item.bootstrap.scss
```diff
@@ -0,0 +1,22 @@
+@use './list-item.base.scss';
```
simeonoff, line 1: This should rather be imported and used as the default static styles in the list-item.ts. It shouldn't be imported here.

src/components/list/themes/light/list-item.fluent.scss
```diff
@@ -1,17 +1,18 @@
+@use './list-item.base.scss';
```
simeonoff, line 1: This should rather be imported and used as the default static styles on the list-header.ts. It shouldn't be imported here.

src/components/list/themes/light/list-item.indigo.scss
```diff
@@ -1,23 +1,19 @@
+@use './list-item.base.scss';
```
simeonoff, line 1: This should rather be imported and used as the default static styles on the list-header.ts. It shouldn't be imported here.

simeonoff (changes requested):

PR #276, opened 2022-03-09
style(switch): update fluent switch styles

### LINKED ISSUE #279: Update switch styles for Fluent theme
The enabled switch when off should use gray 700 for the track border instead of 500

The enabled switch when on should use gray 50 for the thumb instead of surface

The disabled switch when on should use gray 100 for the thumb instead of surface

didimmova:
Closes #279


StefanIvanov (dismissed):

StefanIvanov:
All looks great I only cannot see the focus rectangular border around the switch visual
Screenshot 2022-03-09 at 17 11 51

StefanIvanov (dismissed):

simeonoff (approved):

PR #299, opened 2022-03-18
style(checkbox/switch/radio): add focus outline for fluent theme

didimmova:


StefanIvanov (approved):

StefanIvanov (dismissed):

simeonoff:
@StefanIvanov @didimmova This PR doesn't introduce the requested changes in the issue.

simeonoff (changes requested):
Doesn't seem to fix the issue.

StefanIvanov:
> 



> Doesn't seem to fix the issue.

It seemed to work initially but then @didimmova switched focus-within to focus-visible and not it seems like I am not even able to focus the element at all

simeonoff (approved):

src/components/checkbox/checkbox-base.ts
```diff
@@ -42,6 +42,10 @@ export class IgcCheckboxBaseComponent extends EventEmitterMixin<
   @property({ type: Boolean, reflect: true })
   public invalid = false;
 
+  /** Controls the keyboard focus of the element. */
+  @property({ type: Boolean, reflect: true })
+  public focused = false;
```
rkaraivanov, line 47: Can we make this an internal `@state` property and set the part attribute somewhere in the shadow DOM?
Can you achieve the styling with that approach?
Not really into exposing this as a public property which handles internal state...

rkaraivanov (changes requested):

src/components/checkbox/checkbox-base.ts
simeonoff: Done.

simeonoff (approved):

rkaraivanov (approved):

PR #303, opened 2022-03-25
feat(calendar): Set active date through attribute

### LINKED ISSUE #285: igc-calendar active date should support being set via an attribute like value and values
## Description  
igc-calendar active date should support being set via an attribute like value and values

## Steps to reproduce  

1. attempt to set active-date via an attribute

## Result  
the active date is not mutated.

## Expected result  
the active date should be set in the same manner as if value/values was set via an attribute.

rkaraivanov:
Closes #285


rkaraivanov:
@simeonoff @StefanIvanov 
Currently, the calendar does not apply a distinct style for the ***active date*** property in the days-view component.
Can you give opinion of whether there should be such a style or we can disregard it?

igdmdimitrov (approved):

PR #306, opened 2022-03-31
theme(fluent): update fluent palette primary and secondary colors

### LINKED ISSUE #305: Make palettes consistent
I noticed that the bootstrap dark palette is setting gray and surface colors that are not up to date with recent implementations and definitions, please refer to [this table](https://infragistics.visualstudio.com/Design-System-Model/_wiki/wikis/Design-System-Model.wiki/562/Palettes)

We also decided to introduce a fluent default palette and use it for both light and dark schemas, preserving the word and excel palettes, but using the new one as a default for both, you can also find it in [the same table](https://infragistics.visualstudio.com/Design-System-Model/_wiki/wikis/Design-System-Model.wiki/562/Palettes)

Please check all other palettes against the table in case I have missed anything when reviewing them.

didimmova:
Closes #305


StefanIvanov:
@didimmova , do you plan to update bootstrap-related requests I made in the issue in this PR as well?

StefanIvanov:
Something seems to be off with bootstrap dark theme.
Screenshot 2022-03-31 at 13 51 04

Looking at the _get-shade function I notice that it makes its mind for dar/light based on the luminance of the base gray color and since the folks at BS define it as a value #adb5bd its luminance will always return a value <0.5 so the grays will not generate correctly for dark theme with the value from this PR. However, reverting it to #fff is not correct either. How can we turn things around so that the palette appears correctly?

option 1: change the _get-shade so that it takes a param for dark/light theme and keep gray at #adb5bd

option 2: change the gray to #ced4da so that it reflects 400 variant by Bootstrap and we also have a value that returns correct luminance value for a dark them to look as expected?

simeonoff:
> Something seems to be off with bootstrap dark theme. Screenshot 2022-03-31 at 13 51 04
> 
> Looking at the _get-shade function I notice that it makes its mind for dar/light based on the luminance of the base gray color and since the folks at BS define it as a value #adb5bd its luminance will always return a value <0.5 so the grays will not generate correctly for dark theme with the value from this PR. However, reverting it to #fff is not correct either. How can we turn things around so that the palette appears correctly?
> 
> option 1: change the _get-shade so that it takes a param for dark/light theme and keep gray at #adb5bd
> 
> option 2: change the gray to #ced4da so that it reflects 400 variant by Bootstrap and we also have a value that returns correct luminance value for a dark them to look as expected?
> 
> @simeonoff I hope this is easy to understand :)

I switched to measuring the luminance of the `surface` color instead to decide how the shades of gray should be generated. This fixes the problem.

simeonoff (approved):

PR #314, opened 2022-04-03
Add DateTime Input component

### LINKED ISSUE #247: Date input component

- [ ] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

onlyexeption:
Closes #247


damyanpetev:
Want to put forward for discussion the name of the component - Right now it's Date Input, but it technically supports time portions and should also be able to run in a time-only mask if so configured. We have this as `igxDateTimeEditor` in Angular and it's the base of both a Date Picker and Time Picker. The latter are separate due to the specific picked UI, but the input base drives the same. Just wondering if we want to keep the naming scheme the same before it's too late to do a simple rename and make this a DateTime Input.
Technically, in JS the `Date` is the equivalent of `DateTime` on other platforms so even the shorter name is not entirely unclear, just wondering if a DateTime name would sit better next to a date/time pickers and an eventual date-time picker if we ever do one. Might be more discoverable too, dunno 

//CC @ChronosSF @rkaraivanov 
not sure if I should tag anyone else

ChronosSF:
Yes, let's rename it to igcDateTimeEditor.

ChronosSF (changes requested):
We need to rename this to igcDateTimeEditor

src/components/input/input-base.ts
```diff
@@ -41,8 +40,7 @@ export abstract class IgcInputBaseComponent
   protected inputId = `input-${nextId++}`;
 
   /** The value attribute of the control. */
-  @blazorSuppress()
-  public abstract value: string;
+  //public abstract value: string;
```
damyanpetev, line 43: I guess you dropped this due to the type changing in the date input, though that can be resolved as well - you just need the base to be a superset of all types and the extending classes to narrow it down.

Also, even like this the even mapping is still coming from the base and the `igcChange` interface is still typed as a string and we need to override that. If we make the base generic on the value type it could likely solve that but still need to check.

Also not sure how that'd affect Blazor analysis

src/components/input/input-base.ts
```diff
@@ -110,12 +108,13 @@ export abstract class IgcInputBaseComponent
...
   protected abstract renderInput(): TemplateResult;
 
+  //Remove from here since we wont have value prop in input-base
   protected resolvePartNames(base: string) {
     return {
       [base]: true,
       prefixed: this.prefixes.length > 0,
       suffixed: this.suffixes.length > 0,
-      filled: !!this.value,
+      //filled: !!this.value,
```
damyanpetev, line 117: I don't see this added back in extending classes though, I'm assuming its added for a reason, right?

src/index.ts
```diff
@@ -72,3 +73,4 @@ export { IgcInputBaseComponent } from './components/input/input-base.js';
 export { IgcSliderBaseComponent } from './components/slider/slider-base.js';
 export { IgcButtonBaseComponent } from './components/button/button-base.js';
 export { IgcCalendarBaseComponent } from './components/calendar/common/calendar-base.js';
+export { IgcMaskInputBaseComponent } from './components/mask-input/mask-input-base.js';
```
damyanpetev, line 76: That's kind of odd we have to do this, but sure

damyanpetev:
Something we've discussed was the event map for the mixin that still leaves the `igcChange` event typed as string and how we can possibly update that. It seems we can't use my initial suggestion, since generic arguments can't be used in the mixin expressions ([by design](https://github.com/microsoft/TypeScript/issues/26154)). So an alternative would be for the date input to override the event mixin again to set the types correctly. Something like:
```ts
 export interface IgcDateTimeInputEventMap extends Omit<IgcInputEventMap, 'igcChange'> {
  igcChange: CustomEvent<Date>;
}

export default class IgcDateTimeInputComponent 
  extends EventEmitterMixin<IgcDateTimeInputEventMap, Constructor<IgcInputBaseComponent>>(IgcInputBaseComponent)
```
For this we'll need either an `AbstractConstructor` type or to drop the abstract from the base class ofc.

And then you can make the base `value` the union of possible values and have the extending classes override with the specific one. That actually seems to be done like that already.

In that case must also ping @gmurray81:
- Would a base class property type union mess with analysis if the extending classes narrow it down?
- Same thing for additional mixins applied to extending components

ChronosSF:
Issues I've found so far:

1. [ ] In storybook set displayFormat to d/M/yy h:m:s . Set the date to May 1st from the value calendar. Observe the date input:
Screenshot 2022-05-13 172428
it looks like the displayFormat becomes dd/MM/yy instead (using 'short' works as intended to strip the 0-s btw)

2. [ ] It's not particularly obvious that minValue and maxValue only accept partial/full ISO string. You should make it more clear in the spec and docs for sure

3. [ ] The spin buttons are very wonky - Use a full input format like - MM/dd/yy hh:mm:ss and try to use them on each portion of the date. The result depends on both the portion you have the cursor on and also on the value of the `spinLoop` property. In any case almost none of the interactions work properly. Some examples. `spinLoop` false makes the minutes go down but not up, months to not work at all; `spinLoop`: true makes using down on months to change the day, while up works properly... There seems to be some very general issue with this.

4. [ ] The default state of the input throws warnings in the console. Just observe it after opening the storybook sample:

Screenshot 2022-05-13 194151

ChronosSF:
It seems that we won't have union types because of Blazor so the value type concern can be left out.

I still don't know why these warnings are thrown in Storybook but Dimitar assures me they only appear there.

The display issue seems solved. 

I am marking this verifying on the basis of the current spec.

gmurray81 (dismissed):

src/components/mask-input/mask-input-base.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  @state()
+  protected hasFocus = false;
+
+  @state()
+  protected maskedValue = '';
+
+  /** The mask pattern to apply on the input. */
+  @property()
+  public mask!: string;
```
damyanpetev, line 22: Ahem, I thought the point of having the mask base was to reuse the functionality without inheriting all the public props of the mask that don't make sense to a date input.

In that regard - the `mask` doesn't fit. We should keep a protected backing prop in the base for functionality and only expose public get/set on the mask input

src/components/mask-input/mask-input-base.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  /** The prompt symbol to use for unfilled parts of the mask. */
+  @property()
+  public prompt!: string;
+
+  /** The direction attribute of the control. */
+  @property({ reflect: true })
+  public override dir: 'ltr' | 'rtl' | 'auto' = 'auto';
```
damyanpetev, line 30: There's a type `Direction` exposed on the input - not sure we care to reuse it. Or if we should be reusing this across all components even. Might be a useful indicator for further generated API to reuse the same type as well.

Also if every type of input will be overriding this separately, perhaps we should just move the dir to the input base?
// CC @rkaraivanov

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,654 @@
...
+  public displayFormat!: string;
+
+  @property({ attribute: false })
+  public spinDelta!: DatePartDeltas;
+
+  @property({ type: Boolean, attribute: 'spin-loop' })
+  public spinLoop = true;
+
+  @property({ reflect: true })
+  public override dir: 'ltr' | 'rtl' | 'auto' = 'auto';
```
damyanpetev, line 100: This is already inherited, right?

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,654 @@
...
+
+  @property({
+    converter: (value) => {
+      if (!value) {
+        return null;
+      }
+      return DateTimeUtil.parseIsoDate(value);
+    },
+  })
+  public get value(): Date | null {
```
damyanpetev, line 53: I believe we add `@blazorTwoWayBind('igcChange', 'detail')` to this similar to how it's in the input, though I don't think it'll work correctly while the detail type is still `string`

ChronosSF:
@onlyexeption , I think the only thing I can see not working is the readOnly: true still allows you to enter edit mode and change the value of the editor, just doing so with a disabled-like CSS being applied.

src/components/input/input-base.ts
```diff
@@ -41,8 +40,7 @@ export abstract class IgcInputBaseComponent
   protected inputId = `input-${nextId++}`;
 
   /** The value attribute of the control. */
-  @blazorSuppress()
```
damyanpetev, line 44: Probably not a good idea to drop the suppress here btw.

Although, in the interest of cleaner API gen down the line I'm considering an alternative approach - exposing base properties as public only on the theoretical leaf nodes where the type would be fixed and keeping this one protected instead

ChronosSF (dismissed):

ChronosSF:
@onlyexeption , looks good. If you have time, please, update the storybook sample and ensure all property controls are working (rtl isn't for example) and expose one for the spin settings.

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+  DatePart,
+  DateTimeUtil,
+} from './date-util';
+import { blazorTwoWayBind, watch } from '../common/decorators';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { IgcMaskInputBaseComponent } from '../mask-input/mask-input-base';
+import { partNameMap } from '../common/util';
+import { IgcInputEventMap } from '../input/input-base';
+import { EventEmitterMixin } from '../common/mixins/event-emitter';
+import { AbstractConstructor } from '../common/mixins/constructor';
```
rkaraivanov, line 17: Please fix the import paths for ESM resolution.

For concrete imports add the relevant path suffixes `.js`.
For types, you can omit the suffix if you follow the `import type {T} from 'path'`

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+
+export interface IgcDateTimeInputEventMap
+  extends Omit<IgcInputEventMap, 'igcChange'> {
+  igcChange: CustomEvent<Date | null>;
+}
+
+export default class IgcDateTimeInputComponent extends EventEmitterMixin<
```
rkaraivanov, line 24: Add [documentation](https://github.com/IgniteUI/igniteui-webcomponents/wiki/How-to-document-your-component)

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+    year: 1,
+    hours: 1,
+    minutes: 1,
+    seconds: 1,
+  };
+
+  @property({ attribute: 'input-format' })
+  public get inputFormat(): string {
+    return this._inputFormat || this._defaultMask;
+  }
```
rkaraivanov, line 47: Documentation

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+
+  @property({
+    converter: {
+      fromAttribute: (value: string) =>
+        !value ? null : DateTimeUtil.parseIsoDate(value),
+      toAttribute: (value: Date) => value.toISOString(),
+    },
+  })
+  @blazorTwoWayBind('igcChange', 'detail')
+  public get value(): Date | null {
```
rkaraivanov, line 64: Documentation

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+
+  @property({
+    attribute: 'min-value',
+    converter: {
+      fromAttribute: (value: string) =>
+        !value ? null : DateTimeUtil.parseIsoDate(value),
+      toAttribute: (value: Date) => value.toISOString(),
+    },
+  })
+  public minValue!: Date | null;
```
rkaraivanov, line 87: Doc strings

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+
+  @property({
+    attribute: 'max-value',
+    converter: {
+      fromAttribute: (value: string) =>
+        !value ? null : DateTimeUtil.parseIsoDate(value),
+      toAttribute: (value: Date) => value.toISOString(),
+    },
+  })
+  public maxValue!: Date | null;
```
rkaraivanov, line 97: Doc strings

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+  public displayFormat!: string;
+
+  @property({ attribute: false })
+  public spinDelta!: DatePartDeltas;
+
+  @property({ type: Boolean, attribute: 'spin-loop' })
+  public spinLoop = true;
+
+  @property()
+  public locale = 'en';
```
rkaraivanov, line 109: Documentation

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+    this.setFocus();
+  }
+
+  public clear(): void {
+    if (!this.readonly) {
+      this.maskedValue = '';
+      this.value = null;
+      this.setSelectionRange(0, this.input.value.length);
+    }
+  }
```
rkaraivanov, line 308: Since these are public, they should be documented.

Update the specification with signatures since there is mismatch currently.

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+      this.input.checkValidity() && Object.keys(this.validate()!).length === 0
+    );
+  }
+
+  public stepUp(datePart?: DatePart, delta?: number): void {
+    const targetPart = datePart || this.targetDatePart;
+
+    if (!targetPart || this.readonly) {
+      return;
+    }
```
rkaraivanov, line 283: Not sure about preventing an API call if the input is read-only.
`stepUp/Down` should work in disabled/read-only state as it works with the default input.

@damyanpetev 
Your thoughts on this?

src/components/date-time-input/date-time-input.ts
```diff
@@ -0,0 +1,652 @@
...
+
+    const newValue = this.trySpinValue(targetPart, delta);
+    this.value = newValue;
+    this.setFocus();
```
rkaraivanov, line 287: The `setFocus` has nothing to do with focus and more with `setSelectionRange` 😃 
I suppose these methods are used internally and that is why they do more than they should (keeping/modifying state).

IMO you should change them to do only one thing and move the UI state/interactions in protected handlers.

src/components/date-time-input/date-util.ts
```diff
@@ -0,0 +1,856 @@
+import { isDate } from '../calendar/common/utils';
+import { MaskParser } from '../mask-input/mask-parser';
```
rkaraivanov, line 2: Import paths

src/components/date-time-input/date-util.ts
```diff
@@ -0,0 +1,856 @@
...
+  public static getDefaultMask(locale: string): string {
+    locale = locale || DateTimeUtil.DEFAULT_LOCALE;
+    if (
+      !Intl ||
+      !Intl.DateTimeFormat ||
+      !Intl.DateTimeFormat.prototype.formatToParts
+    ) {
+      // TODO: fallback with Intl.format for IE?
+      return DateTimeUtil.DEFAULT_INPUT_FORMAT;
+    }
```
rkaraivanov, line 129: Not really needed since [our supported targets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#browser_compatibility) are compliant.

src/components/mask-input/mask-input-base.ts
```diff
@@ -0,0 +1,148 @@
+import { property, state } from 'lit/decorators.js';
+import { IgcInputBaseComponent } from '../input/input-base';
+import { MaskParser } from './mask-parser';
```
rkaraivanov, line 3: Import paths

src/index.ts
rkaraivanov: 👍 
Don't export it. 
I would vouch to remove the export of the other base classes as well in the future.

rkaraivanov (changes requested):

src/components/date-time-input/date-time-input.ts
damyanpetev: Agreed, it's public API that should not be limited. Seems like this is due to the public method being used directly in event handling. The read only check should be moved there instead.

src/components/date-time-input/date-util.ts
damyanpetev: probably transferred over from Angular's utils as is, but yeah sure - we can drop the check

src/components/input/input-base.ts
damyanpetev: @onlyexeption This should still remain unchanged

ChronosSF (dismissed):

damyanpetev (dismissed):

src/components/common/definitions/defineAllComponents.ts
```diff
@@ -39,6 +39,7 @@ import IgcToastComponent from '../../toast/toast.js';
 import IgcSliderLabelComponent from '../../slider/slider-label.js';
 import { defineComponents } from './defineComponents.js';
 import IgcCircularGradientComponent from '../../progress/circular-gradient.js';
+import IgcDateTimeInputComponent from '../../date-time-input/date-time-input';
```
rkaraivanov, line 42: import path

rkaraivanov (changes requested):
LGTM
Just fix the import path and it is good to go.

ChronosSF (approved):

rkaraivanov (approved):

PR #317, opened 2022-04-05
feat(navdrawer): add pinned navdrawer position

### LINKED ISSUE #311: Add a relative position option (pinned) to the Nav Drawer
Add support for a drawer that sits with the content (pinned / not-modal). Since the drawer includes `position` that is related to the overlay/modal positions it's odd to expose a `pinned` option or something similar since it'll conflict with the position. To this end, I propose an additional option of the `position` itself - for example `relative`
image

This way it's exclusive with the fixed positions with directions and hopefully avoid confusion.
Whatever the naming we go for, keep in mind a potential future option like `relativeBelow` or `pinedBelow`/`pinThreshold` thing.

NOTES:
From what I can tell something like the following pretty much does the trick
```scss
:host([position='relative']) {
    [part='base'] {
        position: relative;
        width: fit-content;
        box-shadow: none;
    }
}
```

The only issue I see is with the content border, IIRC in Angular that is still controlled by the position, even in pinned state. Something to consider about this approach, because with the exclusive position, the border will always be in this place, which is not a huge deal I think

// CC @StefanIvanov @sdimchevski 

- [ ] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

didimmova:
Closes #311


src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -77,6 +77,15 @@
     }
 }
 
+:host([position='relative']) {
+    [part='base'] {
+        position: relative;
+        width: fit-content;
+        height: 100vh;
```
damyanpetev, line 84: The 100% height from the base styling is just fine, so no need for this. The drawer can actually sit under a nav bar commonly and 100vh will mess with that, so let the surrounding layout control the height.

For the stories demo to show correctly you can simply expand the main container like so:
https://github.com/IgniteUI/igniteui-webcomponents/commit/200afb16864330f9c0dfc4919a22ec25484a3763#diff-21d8dd568fac18b25ddeb66244307d0251e27ea735ae64b36ff2ec202297507dR202

damyanpetev, line 84: The 100% height from the base styling is just fine, so no need for this. The drawer can actually sit under a nav bar commonly and 100vh will mess with that, so let the surrounding layout control the height.

For the stories demo to show correctly you can simply expand the main container like so:
https://github.com/IgniteUI/igniteui-webcomponents/commit/200afb16864330f9c0dfc4919a22ec25484a3763#diff-21d8dd568fac18b25ddeb66244307d0251e27ea735ae64b36ff2ec202297507dR202

didimmova: Do you suggest adding a 'mini' property or just adding control in the story?

damyanpetev: That's separate, the link should be directly to the line that sets the story container height and you've added it so it's good now

damyanpetev:
@didimmova The relative drawer still opens up with a box shadow. Guessing that's related to the shadow/border fix, but the relative style no longer overrides:
image

The relative rule is after the base open one, so this happens. But in any case, perhaps we should make the override more specific to match the new box shadow rule anyway.

stories/nav-drawer.stories.ts
```diff
@@ -13,11 +13,11 @@ const metadata = {
...
   argTypes: {
     position: {
-      type: '"start" | "end" | "top" | "bottom"',
+      type: '"start" | "end" | "top" | "bottom" | "relative"',
       description: 'The position of the drawer.',
-      options: ['start', 'end', 'top', 'bottom'],
+      options: ['start', 'end', 'top', 'bottom', 'relative'],
       control: {
-        type: 'inline-radio',
+        type: 'select',
```
damyanpetev, line 20: heh, that happens to me as well, might as well commit it if it keeps changing :)

damyanpetev (approved):
If everyone is okay with this approach (bit of an edge case w/ the border), this LGTM

PR #318, opened 2022-04-06
Add navdrawer backdrop and fix shadow on closed issue

### LINKED ISSUE #309: Nav Drawer drop shadow visible when the drawer is not open
## Description  
When the Nav Drawer is not open, the drop shadow on the respective side is still visible when using `start` or `end` position. Doesn't seem to be visible. It's pretty faint, but it's there.

## Steps to reproduce  

1. Open the Nav Drawer storybook demo
2. Switch close the drawer
3. Switch between `start` and `end` position
4. Observe the demo sides

## Result  
Kind of hard to show in picture without a comparison since it's so light, but here goes:
image
Also fairly sure I see a bit of border there, if I look hard enough. Not sure if the same thing, but similar can be observed in the [samples](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/menus/navigation-drawer#navbar-integration):
image

## Expected result  
The drop shadow and/or border should not be visible when the nav drawer is not open

### LINKED ISSUE #310: Add a built-in backdrop to the Nav Drawer
In the current state the Nav Drawer is strictly an overlay, but there's no backdrop on the rest of the content (visually) and outside click doesn't close the drawer.

I feel like those are common enough to be a good candidate to integrate into the component. To that end:
- A built-in backdrop can be added to the component to overlay the rest of the viewport the drawer doesn't, A `modal` option (or whatever naming is consistent with similar components) can control the color or if the backdrop is used at all. IMO is should be visible by default.
- Can likely utilize a `closeOnOutsideClick` or similar option.

// CC @StefanIvanov @sdimchevski 

- [ ] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

didimmova:
Closes #309 
Closes #310


src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -28,6 +26,36 @@
     [part='mini'] {
         width: rem(60px);
     }
+
+    [part='overlay'] {
+        opacity: .54;
+        background: color(gray, 500);
```
simeonoff, line 32: You can actually pass a third parameter to the color function - the alpha. So instead of using opacity for the entire container, you could just do:

```suggestion
        background: color(gray, 500, .54);
```

src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -28,6 +26,36 @@
...
+        position: fixed;
+        top: 0;
+        bottom: 0;
+        left: 0;
+        right: 0;
```
simeonoff, line 37:
```suggestion
        inset: 0;
```

src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -28,6 +26,36 @@
...
+    background: color(surface, 500);
+    box-shadow: var(--igc-elevation-16);
+    position: fixed;
+    top: 0;
+    bottom: 0;
+}
+
+:host([position='start']) [part='mini'] {
+    border-inline-end: 1px solid color(gray, 900, .14);
+    left: 0;
```
simeonoff, line 53: Make sure to use the appropriate direction-agnostic properties so that RTL is supported out of the box.

```suggestion
    inset-inline-start: 0;
```

src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -28,6 +26,36 @@
...
+}
+
+:host([position='end']) [part='mini'] {
+    border-inline-start: 1px solid color(gray, 900, .14);
+    right: 0;
```
simeonoff, line 58:
```suggestion
    inset-inline-end: 0;
```

simeonoff (changes requested):
Make sure you use `inset-inline-star` and `inset-inline-end` in place of `left` and `right` respectively. In this way we can ensure full RTL compatibility.

simeonoff (approved):

src/components/nav-drawer/styles/nav-drawer.base.scss
```diff
@@ -68,23 +92,46 @@
...
     [part='main'] {
         display: none;
     }
+
+    [part='overlay'] {
+        display: none;
+    }
+}
+
+:host([position='relative']) {
```
damyanpetev, line 125: Should this be part of this PR?

didimmova: No, I fixed it.

simeonoff (approved):

PR #332, opened 2022-04-13
Add expansion panel component

### LINKED ISSUE #177: Expansion panel component
_Estimated Delivery Sprint TBD

- [x] Research
- [x] Specification
- [ ] API Documentation
- [x] Design implementation - https://www.figma.com/file/GjfbpwcQ1Gh1XEsNwmhHMG/Expansion-Panel?node-id=0%3A1

ddaribo:
Closes #177


src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+ * @csspart headerText
+ * @csspart indicator
+ * @csspart content
+ */
+@themes({ bootstrap, fluent, indigo })
+export default class IgcExpansionPanelComponent extends EventEmitterMixin<
+  IgcExpansionPanelComponentEventMap,
+  Constructor<LitElement>
+>(LitElement) {
+  public static tagName = 'igc-expansion-panel';
```
onlyexeption, line 43: tagName should be readonly

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+          >
+          </igc-icon>
+        </slot>
+      </div>
+    `;
+  }
+
+  private headerTemplate() {
+    return html`
+      <div
```
damyanpetev, line 196: The ARIA design pattern wraps the button in a heading (must check if that's always reqruired), which this is missing. Also if we wrap it internally there's no control over the aria level which might be a bit of an issue.

damyanpetev: Additionally, in terms of accessibility - if there's no title/sub projected this will still render as an empty button that is still targetable by tab navigation which I'm not sure is ideal.

Speaking of, we used expansion panel in some scenarios without a header even and if such a scenario is also intended to be supported then the hardcoded header would not fit.

Tagging @sdimchevski as well for this and I'll look for the example without a header to add here so we can evaluate

damyanpetev: Okay took a bit to dig up since this is (no longer?) embedded in the docs, but the first card in this sample is using an expansion panel:
https://www.infragistics.com/angular-demos/layouts/card-sample-4
Point being, it doesn't have a header and it's toggled by some external button that either need to be positioned by the Card layout or under the panel as in this case. So my question is would this be a case we should consider supporting?

sdimchevski: @damyanpetev I don't know why this Card in the example is using an expansion panel. Expansion panels by design should have a header that acts as the control for expanding and collapsing. I believe you can have a card and implement your own collapsible mechanic through the use of button events but that's not in the scope of the expansion panel.

mmart1n:
The direction property is not bind to the expansion panel within expansion-panel.stories.ts

stories/expansion-panel.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+      <div slot="title">The Expendables</div>
+      <div slot="subTitle">Action, Adventure, Thriller</div>
+      <div>
+        Barney Ross leads the "Expendables", a band of highly skilled
+        mercenaries including knife enthusiast Lee Christmas, martial arts
+        expert Yin Yang, heavy weapons specialist Hale Caesar, demolitionist
+        Toll Road and loose-cannon sniper Gunner Jensen.
+      </div>
+    </igc-expansion-panel>
+    <p style="visibility: collapse">"${direction}"</p>
```
mmart1n, line 83: I believe that this line could be removed.

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
+import { LitElement, html } from 'lit';
+import { property, query } from 'lit/decorators.js';
+import { Constructor } from '../common/mixins/constructor';
+import { EventEmitterMixin } from '../common/mixins/event-emitter';
```
rkaraivanov, line 4:
```suggestion
import { Constructor } from '../common/mixins/constructor.js';
import { EventEmitterMixin } from '../common/mixins/event-emitter.js';
```
Once TypeScript 4.7 hits official, we will try to enforce ESM rules for the package. But currently this will break under Webpack 5

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * @slot title - renders the title of the panel's header
+ * @slot subTitle - renders the subtitle of the panel's header
+ * @slot indicator - renders the expand/collapsed indicator
+ *
+ * @csspart header
+ * @csspart title
+ * @csspart subTitle
+ * @csspart indicator
+ * @csspart content
```
rkaraivanov, line 37: Add summary info for the for the exposed CSS parts

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+ * The Expansion Panel Component provides a way to display information in a toggleable way -
+ * compact summary view containing title and description and expanded detail view containing
+ * additional content to the summary header.
+ *
+ * @element igc-expansion-panel
+ *
+ * @slot title - renders the title of the panel's header
+ * @slot subTitle - renders the subtitle of the panel's header
+ * @slot indicator - renders the expand/collapsed indicator
+ *
```
rkaraivanov, line 32: Add the **@fires [event] - summary** here

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+  public indicatorAlignment: 'start' | 'end' = 'start';
+
+  @query('[part~="header"]', true)
+  protected panelHeader!: HTMLElement;
+
+  private panelId!: string;
+
+  constructor() {
+    super();
+  }
```
rkaraivanov, line 66: Redundant if no additional logic is executed here.

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,239 @@
...
+    super.connectedCallback();
+    const id = this.getAttribute('id');
+    this.panelId! = id ? id : 'igc-expansion-panel-' + ++NEXT_ID;
+  }
+
+  private handleClicked(event: Event) {
+    if (this.disabled) {
+      event.preventDefault();
+      return;
+    }
```
rkaraivanov, line 78: **Suggestion**: IMO this will be unnecessary if you apply `pointer-events: none` on the header part when the control is disabled.

rkaraivanov (changes requested):
LGTM with some minor changes to be applied.

src/index.ts
```diff
@@ -16,6 +16,7 @@ export { default as IgcDropdownComponent } from './components/dropdown/dropdown.
 export { default as IgcDropdownGroupComponent } from './components/dropdown/dropdown-group.js';
 export { default as IgcDropdownHeaderComponent } from './components/dropdown/dropdown-header.js';
 export { default as IgcDropdownItemComponent } from './components/dropdown/dropdown-item.js';
+export { default as IgcExpansionPanelComponent } from './components/expansion-panel/expansion-panel';
```
rkaraivanov, line 19:
```suggestion
export { default as IgcExpansionPanelComponent } from './components/expansion-panel/expansion-panel.js';
```

rkaraivanov (changes requested):

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,235 @@
...
+  @property({ reflect: true, type: Boolean })
+  public open = false;
+
+  /** Get/Set whether the expansion panel is disabled. Disabled panels are ignored for user interactions. */
+  @property({ reflect: true, type: Boolean })
+  public disabled = false;
+
+  /** The indicator alignment of the expansion panel. */
+  @property({ reflect: true, attribute: 'indicator-alignment' })
+  public indicatorAlignment: 'start' | 'end' = 'start';
```
damyanpetev, line 62: Can we rename this to `indicatorPosition`? I think it fits better with what it does

damyanpetev:
@simeonoff Is it me or is the icon a bit smaller than the design?
image
Also assuming the slight difference in perceived subtitle weight is just the specific font being used?

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,235 @@
...
+import { styles as bootstrap } from './themes/light/expansion-panel.bootstrap.css.js';
+import { styles as fluent } from './themes/light/expansion-panel.fluent.css.js';
+import { styles as indigo } from './themes/light/expansion-panel.indigo.css.js';
+
+let NEXT_ID = 0;
+export interface IgcExpansionPanelComponentEventMap {
+  igcOpening: CustomEvent<any>;
+  igcOpened: CustomEvent<any>;
+  igcClosing: CustomEvent<any>;
+  igcClosed: CustomEvent<any>;
```
rkaraivanov, line 16: Please provide the correct type for the custom event `detail` property, since it is not `any`.

rkaraivanov (dismissed):

simeonoff:
> @simeonoff Is it me or is the icon a bit smaller than the design? image Also assuming the slight difference in perceived subtitle weight is just the specific font being used?

The size of the integrated indicator should be fixed now. You are right, the font differences are because of the difference in type face.

simeonoff (approved):

rkaraivanov (approved):

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,235 @@
...
+    const id = this.getAttribute('id');
+    this.panelId! = id ? id : 'igc-expansion-panel-' + ++NEXT_ID;
+  }
+
+  private handleClicked(event: Event) {
+    const el = event.target as HTMLElement;
+
+    if (!el.matches(TABBABLE_SELECTORS)) {
+      this.panelHeader!.focus();
+    }
```
rkaraivanov, line 80: @damyanpetev raised a valid point over this code here.

What is the behavior that we try to achieve here? Shifting focus over to the header?
Isn't the header by default focusable if not disabled?

rkaraivanov (changes requested):

src/components/expansion-panel/expansion-panel.ts
ddaribo: We introduced this because when clicking on non-focusable elements, such as divs containing the title/subtitle slots and the text itself, the header does not seem to take focus. Consequently, as there is no focus after clicking in this case, keyboard navigation does not function properly.

simeonoff (dismissed):

src/components/expansion-panel/expansion-panel.ts
damyanpetev: Seems to be something to do with the shadow DOM handling, otherwise in light DOM a _focusable_ element (that includes `div` with `tabindex`) will be [focused on click as well](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#try_it).
I'm okay with the functionality staying the way it is, though the `TABBABLE_SELECTORS` approach is shaky - the list is far from exhaustive and we have our own components that are direct native replacements (eg `igc-button`). Doesn't seem like feasible/worth it chasing all possible elements. 
Considering the design of the component doesn't really cater to interactive content inside the header, I'm thinking we can probably do with an unconditional focus and leave interactive content to control handling by stopping propagation if needed.

rkaraivanov: @damyanpetev 
Agreed. With the current approach we handle native controls but if for whatever reason I place a `igc-input` component
in the header we don't handle it. Replace with any other custom input. I'm inclined for an unconditional focus on click. That is users can handle the click event in the Light DOM  and stop propagation depending on the scenario rather than do partial handling.

src/components/expansion-panel/expansion-panel.ts
```diff
@@ -0,0 +1,228 @@
...
+  @property({ reflect: true, type: Boolean })
+  public open = false;
+
+  /** Get/Set whether the expansion panel is disabled. Disabled panels are ignored for user interactions. */
+  @property({ reflect: true, type: Boolean })
+  public disabled = false;
+
+  /** The indicator position of the expansion panel. */
+  @property({ reflect: true, attribute: 'indicator-position' })
+  public indicatorPosition: 'start' | 'end' = 'start';
```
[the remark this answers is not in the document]
teodosiah: As far as I remember, we've decided that we won't have None state because there is a slot that could be left empty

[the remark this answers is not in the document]
simeonoff: @teodosiah Providing an empty container feels more like a hack than an OOTB solution.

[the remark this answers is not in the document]
teodosiah: I think it could be also hidden through the styles

simeonoff (approved):

rkaraivanov (approved):

ChronosSF:
so who's going to remove the DO NOT MERGE label

src/components/expansion-panel/themes/light/expansion-panel.bootstrap.scss
```diff
@@ -0,0 +1,10 @@
+@use '../../../../styles/utilities/index' as *;
+
+[part='title'],
+[part~='indicator'] {
+    color: color(gray, 800);
+}
+
+[part='subtitle'] {
+    color: color(gray, 600);
+}
```
YoannaIvanova, line 10: The description text in the Bootstrap theme should use 14px Body 2, not Subtitle 2.

YoannaIvanova: And the content (body) text should use 16px Body 1, not 14px Body 2.

YoannaIvanova (changes requested):

YoannaIvanova (approved):

rkaraivanov (approved):

PR #334, opened 2022-04-14
feat(chip): add prefix and suffix slots

damyanpetev:
Added `prefix` and `suffix` slots as described in the component doc comment.
Deprecated the existing `start` and `end`. Didn't see a decent way to mark that in the template other than a `<!-- deprecated -->` which renders in the final output and decided to skip, let me know if we should add such comments.


simeonoff (approved):

damyanpetev:
[Spec (Draft) is updated](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Chip-Specification-DRAFT#slots)

PR #341, opened 2022-04-20
Add tabs component

### LINKED ISSUE #185: Tab component
_Estimated Delivery Sprint - TBD

- [ ] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

mtsvyatkova:
Closes #185

### DIFF
--- a/src/components/tabs/tab.ts
+++ b/src/components/tabs/tab.ts
@@ -1,0 +1,40 @@
+import { html, LitElement } from 'lit';
+import { property } from 'lit/decorators.js';
+import { styles } from './themes/light/tab.base.css';
+
+export default class IgcTabComponent extends LitElement {
+  public static readonly tagName = 'igc-tab';
+
+  public static override styles = styles;
+
+  @property({ type: String })
+  public panel = '';
+
+  @property({ type: Boolean })
+  public selected = false;
+
+  @property({ type: Boolean, reflect: true })
+  public disabled = false;
+
+  protected override render() {
+    return html`
+      <div
+        part="base"
+        role="tab"
+        aria-disabled=${this.disabled ? 'true' : 'false'}
+        aria-selected=${this.selected ? 'true' : 'false'}
+        tabindex=${this.disabled || !this.selected ? '-1' : '0'}
+      >
+        <slot name="prefix" part="prefix"></slot>
+        <slot></slot>
+        <slot name="suffix" part="suffix"></slot>
+      </div>
+    `;
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'igc-tab': IgcTabComponent;
+  }
+}
--- a/src/components/tabs/tabs.ts
+++ b/src/components/tabs/tabs.ts
@@ -1,0 +1,139 @@
+import '../button/icon-button';
+import { html, LitElement } from 'lit';
+import {
+  property,
+  query,
+  queryAssignedElements,
+  state,
+} from 'lit/decorators.js';
+import { watch } from '../common/decorators';
+import IgcTabComponent from './tab';
+import IgcTabPanelComponent from './tab-panel';
+import { styles } from './themes/light/tabs.base.css';
+
+export default class IgcTabsComponent extends LitElement {
+  public static readonly tagName = 'igc-tabs';
+
+  public static override styles = styles;
+
+  @queryAssignedElements({ selector: 'igc-tab' })
+  private tabs!: Array<IgcTabComponent>;
+
+  @queryAssignedElements({ slot: 'panel' })
+  private panels!: Array<IgcTabPanelComponent>;
+
+  @query('[part="headers-container"]')
+  private headersContainer!: HTMLLIElement;
+
+  @query('[part="selected-indicator"]', true)
+  private selectedIndicator!: HTMLElement;
+
+  @state() private hasLeftScrollButton = false;
+  @state() private hasRightScrollButton = false;
+
+  @property({ type: String })
+  public selected = '';
+
+  @property({ reflect: true })
+  public alignment: 'start' | 'end' | 'center' | 'justify' = 'start';
+
+  @property()
+  public activation: 'focus' | 'select' = 'select';
+
+  protected override firstUpdated() {
+    if (!this.selected) {
+      const firstTab = this.tabs.find((el) => !el.disabled);
+
+      this.setSelectedTab(firstTab?.panel);
+    } else {
+      this.setSelectedTab(this.selected);
+    }
+
+    this.updateScrollButtons();
+  }
+
+  private setSelectedTab(tab?: string) {
+    if (!tab) {
+      this.hideSelectedIndicator();
+      return;
+    }
+
+    if (tab !== this.selected) {
+      this.selected = tab;
+    }
+
+    this.tabs.forEach((el) => (el.selected = el.panel === this.selected));
+    this.panels.forEach((el) => (el.selected = el.name === this.selected));
+
+    this.realignSelectedIndicator();
+  }
+
+  @watch('alignment', { waitUntilFirstUpdate: true })
+  private realignSelectedIndicator() {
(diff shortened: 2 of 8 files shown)


src/components/tabs/tabs.ts
```diff
@@ -0,0 +1,391 @@
...
+        if (
+          tab.offsetLeft + tab.offsetWidth >
+          this.headersWrapper.offsetWidth - this.offset
+        ) {
+          this.scrollElement(tab, false);
+        }
+      }
+    } else {
+      this.hideSelectedIndicator();
+    }
```
rkaraivanov, line 286: Same as `updateScrollPosition`

src/components/tabs/tabs.ts
```diff
@@ -0,0 +1,391 @@
...
+              <igc-icon-button
+                part="end-scroll-button"
+                size="medium"
+                variant="flat"
+                name="navigate_next"
+                collection="internal"
+                @click=${this.handleEndButtonClick}
+              ></igc-icon-button>
+            `
+          : ''}
```
rkaraivanov, line 378: Consider extracting the scroll button rendering into a separate renderer which takes the condition and the props.
Then this could be reduced to:
```html
<div part="headers">
    ${this.renderScrollButton(condition, {...props})}
    <div part="headers-content">
        <div part="headers-wrapper">
          <div part="headers-scroll" role="tablist">
            <slot></slot>
          </div>
          <div part="selected-indicator"></div>
        </div>
    </div>
    <!-- OR ${this.renderHeaders() } -->
    ${this.renderScrollButton(condition, {...props})}
</div>
```

rkaraivanov (changes requested):

src/components/tabs/themes/light/tab.base.scss
```diff
@@ -0,0 +1,74 @@
...
+    background: color(surface, 500);
+    color: color(gray, 700);
+
+    > * {
+        margin-inline-start: rem(12px);
+
+        &:first-child {
+            margin-inline-start: 0;
+        }
+    }
```
simeonoff, line 30: We can replace all this with:
```suggestion
    gap: rem(12px);
```

src/components/tabs/themes/light/tab.base.scss
```diff
@@ -0,0 +1,74 @@
...
+    outline-style: none;
+
+    ::slotted(*) {
+        @include line-clamp(2, true, true);
+    }
+}
+
+::slotted(igc-icon) {
+    width: rem(24px);
+    height: rem(24px);
```
simeonoff, line 73: Use `--size` for sizing `igc-icon`s.

simeonoff (changes requested):
I can see a slight border leftover in the selected tab when using bootstrap.

rkaraivanov:
@mtsvyatkova @SisIvanova 
Some issues while playing with the component:

* The indicator gets misaligned when clicking on the scroll buttons

https://user-images.githubusercontent.com/20106115/169240688-c61fc130-5e90-453e-8b5f-3823a51b38bf.mp4

* There is a double hover in the fluent theme for the scroll button

https://user-images.githubusercontent.com/20106115/169240827-31937e36-6c40-4bb5-982b-6bd5e7c8ecad.mp4

YoannaIvanova (changes requested):
@mtsvyatkova 
When scrolling tabs right, the left header button is missing.

YoannaIvanova:
@SisIvanova 
- There should be 8px padding between the icon and the text label (all themes);
- Tabs with text labels should have 48px height (bootstrap, material, indigo) and 42px height in fluent;
- The font size in Indigo should be 12px instead of 14px;
- The font size in Bootstrap should be 16px (medium font-weight) instead of 14.4px;

src/components/tabs/tab.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.tab.blur();
+  }
+
+  protected override render() {
+    return html`
+      <div
+        part="base"
+        role="tab"
+        aria-disabled=${this.disabled ? 'true' : 'false'}
+        aria-selected=${this.selected ? 'true' : 'false'}
```
damyanpetev, line 58: I don't see `aria-controls` linking to the panel the tab controls, the specs say 'should' but I think it might be needed to properly reflect the active tab when the focus is in the panel or its children.
To that end - @rkaraivanov would it not make more sense to have `id` on the panel instead of `name` as the ARIA link is based on id-s anyway?

YoannaIvanova (dismissed):

src/components/tabs/tab.ts
rkaraivanov: @damyanpetev 
Fine by me. Feel free to update the specification.

src/components/tabs/tabs.ts
```diff
@@ -0,0 +1,400 @@
...
+      this.tabs.forEach((tab) => (tab.selected = false));
+      this.panels.forEach((panel) => (panel.selected = false));
+      this._selected = '';
+    }
+
+    this.setupObserver();
+  };
+
+  /** Selects the specified tab and displays the corresponding panel.  */
+  public select(selectedTab: string | HTMLElement) {
```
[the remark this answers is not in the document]
rkaraivanov: AFAIK the **selected** property does not notify the parent about the state change.
Otherwise, I'm ok with the method constrained to just the id of the tab (that is a string).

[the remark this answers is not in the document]
damyanpetev: Oh, well perhaps it should? I mean it's public API and you can set it, right?

mtsvyatkova:
There is an issue with the scroll buttons not being properly displayed in Safari browser. Video attached:

https://user-images.githubusercontent.com/11231206/174297271-b707b18f-c3d2-4613-a76f-fe12cb93f9db.mp4

simeonoff:
> There is an issue with the scroll buttons not being properly displayed in Safari browser. Video attached:
> 
>  TabsSafari.mp4

Hm, I can't seem to reproduce this in Safari 15.5. However, another issue is present. Pressing the arrow buttons does nothing in Safari 15.5 and Safari Technology Preview and doesn't work correctly in Chrome.

src/components/tabs/tabs.ts
```diff
@@ -0,0 +1,418 @@
...
+      this.tabs.forEach((tab) => (tab.selected = false));
+      this.panels.forEach((panel) => (panel.style.display = 'none'));
+      this._selected = '';
+    }
+
+    this.setupObserver();
+  };
+
+  /** Selects the specified tab and displays the corresponding panel.  */
+  public select(selectedTab: string | HTMLElement) {
```
damyanpetev, line 362: Still want to perhaps drop the union here, however I checked the generated output and at least for now it picks the string only, so that's perhaps fine

src/components/tabs/tab.ts
damyanpetev: I did though it's described as _should_ have.
Reading https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role we might also consider aria-owns list on the tabs themselves it seems as it's described as _must_

damyanpetev (commented):
Other than some small ARIA updates, LGTM

src/components/tabs/tab.ts
damyanpetev: I did though it's described as _should_ have.
Reading https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role we might also consider aria-owns list on the tabs themselves it seems as it's described as _must_

ChronosSF:
> Other than some small ARIA updates, LGTM

Let's log them separately

simeonoff (approved):

simeonoff (approved):

damyanpetev (approved):

rkaraivanov (approved):

damyanpetev (approved):

PR #343, opened 2022-05-03
Add Dialog component

### LINKED ISSUE #175: Dialog component
_Estimated Delivery FEB 22

- [x] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

igdmdimitrov:
Closes #175


src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,270 @@
...
+  /** Sets the return value for the dialog. */
+  @property({ type: String, attribute: 'return-value' })
+  public returnValue!: string;
+
+  @watch('open')
+  protected async handleOpenState() {
+    await this.updateComplete;
+
+    if (this.nativeElement) {
+      const hasOpenAttr = this.nativeElement.hasAttribute('open');
```
rkaraivanov, line 139: What is the purpose of this check here?

YoannaIvanova:
In the Material theme, the title's and content's font family should be Titillium Web, not Roboto.

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,270 @@
...
+  IgcDialogEventMap,
+  Constructor<LitElement>
+>(LitElement) {
+  public static readonly tagName = 'igc-dialog';
+  public static styles = [styles];
+
+  private titleId = `title-${nextId++}`;
+  private _ariaLabel!: string;
+  private _ariaLabelledby!: string;
+  private _ariaDescribedby!: string;
```
rkaraivanov, line 53: The current limitations of referencing nodes between shadow/light DOM in ARIA context, make these kind of useless. I suggest you remove the labelledby and describedby props

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,270 @@
...
+  @property({ reflect: true })
+  public get role() {
+    return this._role;
+  }
+
+  public set role(value: 'dialog' | 'alertdialog') {
+    const oldVal = this._role;
+    this._role = value;
+    this.removeAttributeAndUpdate('role', 'role', oldVal);
+  }
```
rkaraivanov, line 88: Can we set the role attribute in the connectedCallback if not present?
Seems like an overkill?

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,270 @@
...
+        aria-labelledby=${this.ariaLabelledby
+          ? this.ariaLabelledby
+          : this.titleId}
+        aria-describedby=${ifDefined(this.ariaDescribedby)}
+      >
+        <header part="title" id=${this.titleId}>
+          <slot name="title"></slot>
+          ${this._titleElements.length === 0
+            ? html`<span>${this.title}</span>`
+            : ''}
```
rkaraivanov, line 253:
```suggestion
          <slot name="title"><span>${this.title}</span></slot>
```

rkaraivanov: You can also remove the `_titleElements` property then

rkaraivanov (changes requested):

package.json
```diff
@@ -42,6 +42,7 @@
     "format": "eslint --ext .ts . --fix --ignore-path .gitignore && prettier \"**/*.ts\" --write --ignore-path .gitignore",
     "release": "changelog",
     "test": "npm run build && wtr --coverage",
+    "test:component": "npm run build && wtr \"dist/src/components/dialog/*.spec.js\" --coverage",
```
rkaraivanov, line 45: I think this should be removed :)

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,260 @@
...
+  IgcDialogEventMap,
+  Constructor<LitElement>
+>(LitElement) {
+  public static readonly tagName = 'igc-dialog';
+  public static styles = [styles];
+
+  private static readonly increment = createCounter();
+
+  private titleId = `title-${IgcDialogComponent.increment()}`;
+  private _ariaLabel!: string;
```
rkaraivanov, line 52:
```suggestion
```
See `render` for additional info

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,260 @@
...
+  @property({ attribute: 'aria-label' })
+  public override get ariaLabel() {
+    return this._ariaLabel;
+  }
+
+  public override set ariaLabel(value: string) {
+    const oldVal = this._ariaLabel;
+    this._ariaLabel = value;
+    this.removeAttributeAndUpdate('aria-label', 'ariaLabel', oldVal);
+  }
```
rkaraivanov, line 83:
```suggestion
```
See `render` for additional info

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,260 @@
...
+    }
+  }
+
+  public override connectedCallback() {
+    super.connectedCallback();
+
+    if (!this.getAttribute('role')) {
+      this.setAttribute('role', 'dialog');
+    }
+  }
```
rkaraivanov, line 111:
```suggestion
```
See `render` for additional info

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,260 @@
...
+    attribute: string,
+    name: string,
+    oldVal: string
+  ) {
+    if (this.hasAttribute(attribute)) {
+      this.removeAttribute(attribute);
+    }
+
+    this.requestUpdate(name, oldVal);
+  }
```
rkaraivanov, line 162:
```suggestion
```

src/components/dialog/dialog.ts
```diff
@@ -0,0 +1,260 @@
...
+        <section part="content">
+          <slot @slotchange=${this.handleSlotChange}></slot>
+        </section>
+        <footer part="footer">
+          <slot name="footer"></slot>
+        </footer>
+      </dialog>
+    `;
+  }
+}
```
rkaraivanov, line 254:
```suggestion
protected override render() {
    const label = this.ariaLabel ? this.ariaLabel : undefined;
    const labelBy = label ? undefined : this.titleId;

    return html`
      <div part="backdrop" aria-hidden="true" ?hidden=${!this.open}></div>
      <dialog
        part="base"
        @click=${this.handleClick}
        @cancel=${this.handleCancel}
        aria-label=${ifDefined(label)}
        aria-labelledby=${ifDefined(labelBy)}
      >
        <header part="title" id=${this.titleId}>
          <slot name="title"><span>${this.title}</span></slot>
        </header>
        <section part="content">
          <slot @slotchange=${this.handleSlotChange}></slot>
        </section>
        <footer part="footer">
          <slot name="footer"></slot>
        </footer>
      </dialog>
    `;
  }
```
I've spent some time testing the component with NVDA and I came to the following conclusions:

- we don't need the role attribute on the component since the interactable part is the underlying dialog element which implicitly has the correct role.
- in addition, having role=dialog on the component without the rest of the aria attributes is an error.
- if `aria-label` is provided for the igc-dialog, pass it down to the native dialog and remove the labelledby attribute.
- otherwise use the labelledby attribute, since the id reference is in the same DOM tree (`titleID`)

rkaraivanov (changes requested):
@sdimchevski 

After some reading up, [the first point in the article](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#required_javascript_features)
caught my attention.

This begs the question what do we render when the dialog is opened in this scenario?
```html
<igc-dialog title="Warning"></igc-dialog>
```
image

Maybe we should have a default button when no content is provided just to move focus there for easier closing?

sdimchevski:
@rkaraivanov Yes, dialogs must have at least one focusable control so I think it's a good idea to always have an OK button.

src/components/dialog/dialog.spec.ts
```diff
@@ -0,0 +1,346 @@
...
+      await elementUpdated(dialog);
+
+      expect(dialog).dom.to.have.text(content);
+      expect(dialog).dom.to.equal(
+        `
+        <igc-dialog>
+          <span>
+            Dialog content
+          </span>
+        </igc-dialog>`,
```
damyanpetev, line 55: Confused. No way this is actually testing if the dialog has the content since it's not looking in the shadow DOM - so not really checking if the content is inside the dialog for real.
Same line of thinking for the second check - which again tests the light DOM for the very same nodes that were used in `createDialogComponent`, thus testing if the fixture and native DOM node creation works, kinda pointless.

damyanpetev: PS: Tested by deleting the default slot in the dialog, this test still passed :)

damyanpetev (approved):
Outside of a minor nit and a questionable test, LGTM

PR #355, opened 2022-05-10
feat(icon-button): allow slotted children

### LINKED ISSUE #346: Allow slot projection for the Icon Button
As the title states - wondering if it'd be possible to add a slot to the icon button for projected content. Seems like it'd be easy enough to wrap even the default icon markup in a slot so projected content can override.

Reasoning: I'd like to able to use alternative icons with our icon button, which is currently locked to the `igc-icon` which in turn is svg only. For simple icons it's easy enough to use alternatives - just use the provided CSS as described by whatever library without the need for our component. But for the icon button that's not an alternative and the normal button that does accept content has different design. What I'd like to be able to do is get the icon button looks (backgrounds, interaction) but provide the icon - be it from a font set or a simple emoji without having to go through the svg register process.

The only hurdle with this suggestion would be any styling that explicitly targets the internal `igc-icon`, haven't checked how much of that is there and if it can be adjusted.

- [ ] Research
- [ ] Specification
- [ ] API Documentation
- [ ] Design implementation

simeonoff:
Closes #346


SisIvanova (dismissed):

damyanpetev:
@simeonoff 
When you set some text emoji as content it messes up the material theme circle:
image
Tried with both text directly and wrapped in a span, so doesn't seem like it's just the slotted styles

simeonoff:
> @simeonoff When you set some text emoji as content it messes up the material theme circle: image Tried with both text directly and wrapped in a span, so doesn't seem like it's just the slotted styles

Should be fine now.

SisIvanova (approved):

damyanpetev (approved):

PR #359, opened 2022-05-10
style(input): add hidden overflow to prefix and sufix

### LINKED ISSUE #353: Input prefix/suffix containers expand outside the main box with larger elements
## Description  
When using Bootstrap theme, projecting larger elements into the prefix/suffix slots break the input layout.

## Steps to reproduce  

1. Run the Input story
2. Set the following as prefix:
  ```html
  <span style="font-size: 20px;" slot="prefix">👀</span>
  ```

## Result  
image
It also inexplicably breaks the suffix container.

## Expected result  
There's still enough space to fit items larger than 20x20, so it can/should look like this:
image

## Attachments  

didimmova:
Closes #353

### DIFF
--- a/src/components/input/themes/light/input.bootstrap.scss
+++ b/src/components/input/themes/light/input.bootstrap.scss
@@ -57,6 +57,7 @@ $input-background: var(--input-background, #fff) !default;
         color: utils.color(gray, 800);
         background: utils.color(gray, 300);
         transition: box-shadow .15s ease-out, border .15s ease-out;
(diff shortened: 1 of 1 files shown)


simeonoff (commented):
Do we need to make this change only in the bootstrap theme? Doesn't the issue appear in all themes? If that's the case, then we may need to make the change in the base theme.

simeonoff (approved):

simeonoff (approved):

PR #386, opened 2022-05-25
build: Updated dependencies

rkaraivanov:
* Fixed converter type errors


rkaraivanov:
@simeonoff 
There is build failure with the newer version of stylelint. Can you check if it an actual issue or a rule that is not applicable in our case?

PR #389, opened 2022-05-26
fix(expansion-panel): not working with headings

simeonoff:
The expansion panel, as it was, was not working with heading elements when passed to the slots for title, subtitle, etc. Additional improvements are introduced in this PR.

### DIFF
--- a/src/components/expansion-panel/themes/light/expansion-panel.base.scss
+++ b/src/components/expansion-panel/themes/light/expansion-panel.base.scss
@@ -40,9 +40,11 @@ $padding: calc(var(--igc-spacing-large) * 2) calc(var(--igc-spacing-large) * 3);
     margin: 0;
 }
 
-[part='subtitle']::slotted(*) {
-    @include type-category('subtitle-2');
-    @include ellipsis();
+[part='subtitle'] {
+    &::slotted(*) {
+        @include type-category('subtitle-2');
+        @include ellipsis();
+    }
 
     color: color(gray, 700);
 }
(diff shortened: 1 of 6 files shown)


rkaraivanov (dismissed):

didimmova (dismissed):

src/components/expansion-panel/themes/light/expansion-panel.base.scss
```diff
@@ -54,11 +54,11 @@ $padding: calc(var(--igc-spacing-large) * 2) calc(var(--igc-spacing-large) * 3);
     padding: $padding;
 }
 
-[part~='indicator'] {
+slot[name='indicator'] {
```
ddaribo, line 57: It seems that custom styles targeted to the exposed parts do not have priority over the default styles, which target the slots.
In the styling sample, the following rule is expected to modify the indicator color:
`
igc-expansion-panel::part(indicator) {
    color: #ffd351;
}
`

However, with the current changes, this is the result:

exp-panel-styles
Same for the other exposed parts and slots (title, subtitle).

simeonoff: @ddaribo should be fine now.

ddaribo: The indicator is fine now.
However, setting the title and subtitle colors via the corresponding parts still does not seem to work, I guess because of the change icluding the ::slotted pseudo-element.

ddaribo (approved):

PR #404, opened 2022-06-02
fix(themes): replace root with universal selector

### LINKED ISSUE #400: Themes: customizing `h`, `s`, or `l`, attributes of a palette color has no effect on component
## Description  
Customizing only one component of a palette color has no effect on color when scoped to a specific component selector.

## Steps to reproduce  

For instance, if we wanted to slightly change the primary color shades in the Icon Button, we can't do it by modifying the `hue` variable.

#### Example:
This doesn't work.
```css
igc-icon-button {
    --igc-primary-h: 128deg;
}
```

Redeclaring the color in question `fixes` the issue, however this doesn't sound like a viable approach:

#### Example:
This works.
```css
igc-icon-button {
   --igc-primary-h: 128deg;
   --igc-primary-500: var(--igc-primary-h), var(--igc-primary-s), var(--igc-primary-l);
}
```

## Expected result  
Modifying only a single element of the color, like `hue`, `saturation`, or `lightness` should update the color when done on a component level.

SisIvanova:
Closes #400


simeonoff (approved):

simeonoff:
Closing as this has to be done in another way.

PR #425, opened 2022-06-27
Feat: Select Component

### LINKED ISSUE #180: Select component
_Estimated Delivery FEB 22

- [x] Research
- [x] Specification
- [ ] API Documentation
- [ ] Design implementation

simeonoff:
Closes #180


SisIvanova (approved):

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+    return !this.invalid;
+  }
+
+  public override async firstUpdated() {
+    super.firstUpdated();
+    await this.updateComplete;
+
+    const selectedItem = this.allItems.find((i) => i.selected) ?? null;
+    if (selectedItem) this.value = selectedItem.value;
+    else this.updateSelected();
```
damyanpetev, line 153: If-else without any brackets, that's a new thing. Can't say I'm a fan of the readability.

Also if there's a selectedItem it should hit the `updateValue` watch below, no? Which in turn can (will?) trigger `updateSelected` since that's on a watch too?

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  @watch('selectedItem')
+  protected updateValue() {
+    this.value = this.selectedItem?.value ?? this.value;
```
damyanpetev, line 158: This looks like it can lead to some odd states, if the value of the item is falsy somehow (I know, highly unlikely but still) and maintain the previous value. Basically, if there's a `selectedItem`, its value goes regardless what it is

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+      activeItemIndex ?? -1,
+      -1
+    );
+    this.selectItem(this.allItems[prev], true);
+  }
+
+  protected selectFirstActiveItem() {
+    const items = this.allItems.filter((i) => !i.disabled);
+    const index = this.allItems.indexOf(items[0]);
+    const item = this.allItems[index];
```
damyanpetev, line 210: Seems like the index is not needed for anything else, so this can be reduced to (also consistent with dd logic):
```suggestion
    const item = this.allItems.filter((item) => !item.disabled).at(0);
```
Same goes for the last item below with `.at(-1)`. Also, on the off chance there are no items or something, do check with `if (item && item.value...)` to be safe

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    const item = this.allItems
+      .filter((i) => !i.disabled)
+      .find((i) =>
+        i.textContent
+          ?.trim()
+          .toLowerCase()
+          .startsWith(this.searchTerm.toLowerCase())
+      );
+    if (item) this.selectItem(item);
```
damyanpetev, line 256: I believe this is intended to be equivalent with other keyboard navigation - i.e. activate, rather than select. I realize that's how the native select does it, but it also doesn't have a separate active state and does the same with arrows.
Might need a spec update

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+      this.hide();
+    }
+
+    this.handleKeyDown(event);
+    this.selectItemByKey(event);
+  }
+
+  protected renderInnerSlots(el: HTMLElement, slot: string) {
+    el.setAttribute('slot', slot);
+    return html`${el.cloneNode(true)}`;
```
damyanpetev, line 318: Must be a way to re-project the prefix and suffix slots without cloning them :( This drops the connection the slotted elements had outside the component's shadow DOM and add limitations can be used. I realize in most cases it's a simple icon, but still.

damyanpetev: Assuming you can't re-project via wrapper since it'll trigger the prefixed/suffixed state of the input and mess it up?
Since the input doesn't have a default slot, perhaps we can use that and project/unproject prefix/suffix only when one is present by manipulating the wrapper slot target?
I tried something similar here: [https://lit.dev/playground](https://lit.dev/playground/#project=W3sibmFtZSI6InNsb3Qtc25pZmZlci50cyIsImNvbnRlbnQiOiJpbXBvcnQge2h0bWwsIGNzcywgTGl0RWxlbWVudH0gZnJvbSAnbGl0JztcbmltcG9ydCB7Y3VzdG9tRWxlbWVudCwgcHJvcGVydHksIHF1ZXJ5QXNzaWduZWRFbGVtZW50cywgc3RhdGUgfSBmcm9tICdsaXQvZGVjb3JhdG9ycy5qcyc7XG5cbkBjdXN0b21FbGVtZW50KCdzbG90LXNuaWZmZXInKVxuZXhwb3J0IGNsYXNzIFNsb3RTbmlmZmVyIGV4dGVuZHMgTGl0RWxlbWVudCB7XG4gIHN0YXRpYyBzdHlsZXMgPSBjc3NgcCB7IGNvbG9yOiBibHVlIH1gO1xuXG4gIEBwcm9wZXJ0eSgpXG4gIG5hbWUgPSAnU29tZWJvZHknO1xuICBcbiAgQHF1ZXJ5QXNzaWduZWRFbGVtZW50cyh7IHNsb3Q6ICduYW1lZCcgfSlcbiAgaXRlbXMhOiBBcnJheTxIVE1MRWxlbWVudD47XG4gIFxuICBAc3RhdGUoKVxuICBwcm90ZWN0ZWQgc2xvdFBvcHVsYXRlZCA9IGZhbHNlO1xuXG4gIHJlbmRlcigpIHtcbiAgICByZXR1cm4gaHRtbGBcbiAgICAgICAgPHA-JHt0aGlzLnNsb3RQb3B1bGF0ZWQgPyBgSSB3aWxsIHJlLXNsb3RgIDogYEknbSBlbXB0eSBpbnNpZGVgfSAoJHt0aGlzLml0ZW1zLmxlbmd0aH0pPC9wPlxuICAgICAgICA8c3BhbiBzbG90PVwiJHt0aGlzLnNsb3RQb3B1bGF0ZWQgPyBcInRhcmdldFwiIDogXCJcIn1cIj5cbiAgICAgICAgICA8c2xvdCBuYW1lPVwibmFtZWRcIiBAc2xvdGNoYW5nZT0ke3RoaXMuaGFuZGxlU2xvdENoYW5nZX0-PC9zbG90PlxuICAgICAgICA8L3NwYW48PlxuICAgIGA7XG4gIH1cblxuICBwcm90ZWN0ZWQgb3ZlcnJpZGUgZmlyc3RVcGRhdGVkKCkge1xuICAgIHRoaXMuaGFuZGxlU2xvdENoYW5nZSgpO1xuICB9XG4gIFxuICBwcm90ZWN0ZWQgaGFuZGxlU2xvdENoYW5nZSgpIHtcbiAgICB0aGlzLnNsb3RQb3B1bGF0ZWQgPSAhIXRoaXMuaXRlbXMubGVuZ3RoO1xuICB9XG59XG4ifSx7Im5hbWUiOiJpbmRleC5odG1sIiwiY29udGVudCI6IjwhRE9DVFlQRSBodG1sPlxuPGhlYWQ-XG4gIDxzY3JpcHQgdHlwZT1cIm1vZHVsZVwiIHNyYz1cIi4vc2xvdC1zbmlmZmVyLmpzXCI-PC9zY3JpcHQ-XG48L2hlYWQ-XG48Ym9keT5cbiAgPHNsb3Qtc25pZmZlcj5cbiAgICA8c3BhbiBzbG90PVwibmFtZWRcIj5pdGVtPC9zcGFuPlxuICA8L3Nsb3Qtc25pZmZlcj5cbjwvYm9keT5cbiJ9LHsibmFtZSI6InBhY2thZ2UuanNvbiIsImNvbnRlbnQiOiJ7XG4gIFwiZGVwZW5kZW5jaWVzXCI6IHtcbiAgICBcImxpdFwiOiBcIl4yLjAuMFwiLFxuICAgIFwiQGxpdC9yZWFjdGl2ZS1lbGVtZW50XCI6IFwiXjEuMC4wXCIsXG4gICAgXCJsaXQtZWxlbWVudFwiOiBcIl4zLjAuMFwiLFxuICAgIFwibGl0LWh0bWxcIjogXCJeMi4wLjBcIlxuICB9XG59IiwiaGlkZGVuIjp0cnVlfV0)
Would this work?

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  protected override render() {
+    return html`
+      <igc-input
```
damyanpetev, line 323: There's quire a bit of ARIA missing, such as the role of the control and the linkage to the listbox and its active item. Then again, currently the `igc-input` will not transfer such props to the native input, right?

src/components/select/select.ts
```diff
@@ -0,0 +1,377 @@
...
+  @query('igc-input')
+  protected override target!: IgcInputComponent;
+
+  /** The value attribute of the control. */
+  @property({ reflect: false, type: String })
+  public value!: string | undefined;
+
+  /** The name attribute of the control. */
+  @property()
+  public name!: string;
```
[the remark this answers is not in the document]
simeonoff: It definitely should not be passed to the igc-input.

src/components/select/select.ts
simeonoff: If there's a selected item it will hit, yes. I will refactor this. This wasn't possible before the dropdown refactoring. The only case that should be handled is when there's no `selectedItem` and `value` has been initialized on the select component.

src/components/select/select-group.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  protected override getParent() {
+    return this.closest('igc-select')!;
+  }
+
+  protected override async firstUpdated() {
+    await this.updateComplete;
+    this.controlledItems = this.activeItems;
+
+    this.items.forEach((i) => {
```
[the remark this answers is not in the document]
simeonoff: This leads to bad UX, though. If the developer sets some children as disabled toggling the disabled state on the group will overwrite their state. We've already had a similar issue in igniteui-angular. Disabled items should maintain their disabled state even after the group's disabled state is removed. This is how the native select works at least. Just my two cents.

src/components/select/select-item.ts
```diff
@@ -0,0 +1,48 @@
...
+  /** Returns the text of the item without the prefix and suffix content. */
+  public override get textContent() {
+    return this.content
+      .map((t) => t.textContent?.trim())
+      .filter((t) => t !== '')
+      .join(' ');
+  }
+
+  /** Sets the textContent of the item without touching the prefix and suffix content. */
+  public override set textContent(value: string) {
```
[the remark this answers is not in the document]
simeonoff: Except the prefix and suffix nodes are part of the textContent otherwise as well. This leads to them being included as input value, which is shown when an item is selected. This is bad UX IMO. Also, setting textContent on an `igc-select-item` will remove all prefix/suffix nodes otherwise. Overriding the textContent methods seemed like the best solution to me.

src/components/select/select-group.ts
```diff
@@ -0,0 +1,90 @@
...
+  public override items!: Array<IgcSelectItemComponent>;
+
+  @queryAssignedElements({
+    flatten: true,
+    selector: 'igc-select-item:not([disabled])',
+  })
+  protected activeItems!: Array<IgcSelectItemComponent>;
+
+  @property({ reflect: true, type: Boolean })
+  public disabled = false;
```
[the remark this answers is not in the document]
simeonoff: @sdimchevski Do you think users should be able to disable all items in a drop-down group?

src/components/select/select.ts
```diff
@@ -0,0 +1,425 @@
...
+    if (item.value !== this.value) {
+      this.selectItem(item, true);
+    }
+  }
+
+  protected searchItem(event: KeyboardEvent): void {
+    if (!event || !event.key || event.key.length > 1 || event.key === ' ') {
+      // ignore longer keys ('Alt', 'ArrowDown', etc) AND spacebar (used of open/close)
+      return;
+    }
```
rkaraivanov, line 263:
```suggestion
    // `/^\p{L}|\p{N}$/u` is event more pedantic
    if (!/^.$/u.test(event.key)) {
      return;
    }
```

rkaraivanov (changes requested):
LGTM

Just some minor nitpicks

src/components/select/select.ts
```diff
@@ -0,0 +1,425 @@
...
+  /** The direction attribute of the control. */
+  @property({ reflect: true })
+  public override dir: 'ltr' | 'rtl' | 'auto' = 'auto';
+
+  constructor() {
+    super();
+    this.size = 'medium';
+
+    /** Return the focus to the target element when closing the list of options. */
+    this.addEventListener('igcClosing', () => this.target.focus());
```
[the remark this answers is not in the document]
simeonoff: Setting this on the igcClosed event fires an extra focus event between the closing and the closed events.

rkaraivanov (approved):

ChronosSF:
Changing the value resets the invalid state set through the API. It probably makes sense but it's better to have the behavior in the spec as well.

ChronosSF:
Method observations:
- normalize the way return types are described in the spec
- navigateTo returns an igc-select-item, not value or index
- reportValidity returns boolean

simeonoff:
> Testing notes:
> 
> 1. Many options in storybook don't work if set through Storybook's UI
> 2. Spec needs a bit of love - autocomplete is listed as supported, but the prop is not available and probably doesn't make sense since select is not editable
> 3. Keyboard nav in the spec is not implemented the same way in the component
>    
>    * alt + left and right also open the dropdown, when in the spec only Alt + up/down should do it
>    * left and right are supposed to navigate the items according to the spec but they don't (they do it while the list is closed but the spec doesn't seem to suggest they only work while the dropdown is closed)
> 
> Still to test: methods, events

1. The options that don't do anything in storybook, unfortunately, are inherited from the dropdown component, but are not wired in storybook. We intend to not talk about those options in the documentation as these options should be obsolete in a future update when the dropdown component is refactored and the non-working options are made private and accessible only from the dropdown class.
2. Agreed.
3. To be updated.

ChronosSF:
Ok, the other things I wanted to test look good, putting verified with the caveat that the spec needs to be updated to reflect the way navigation is implemented.

simeonoff:
> Changing the value resets the invalid state set through the API. It probably makes sense but it's better to have the behavior in the spec as well.

The input component works the same way. Perhaps this requires further discussion but I think we should refactor all input-like components and their validity in another PR when we implement native form support for such elements.

PR #428, opened 2022-07-07
Rating: Add support for single selection and empty symbols

### LINKED ISSUE #4: Rating component

simeonoff:
Closes #4


src/components/rating/rating.ts
```diff
@@ -3,21 +3,22 @@ import {
...
-import { styleMap } from 'lit/directives/style-map.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
+import { styleMap } from 'lit/directives/style-map.js';
+import { themes } from '../../theming/theming-decorator.js';
 import { watch } from '../common/decorators/watch.js';
 import { Constructor } from '../common/mixins/constructor.js';
 import { EventEmitterMixin } from '../common/mixins/event-emitter.js';
 import { SizableMixin } from '../common/mixins/sizable.js';
+import { clamp } from '../common/util.js';
+import IgcRatingSymbolComponent from './rating-symbol';
```
rkaraivanov, line 17:
```suggestion
import IgcRatingSymbolComponent from './rating-symbol.js';
// OR
import type { IgcRatingSymbolComponent  } from './rating-symbol';
```

src/components/rating/rating.ts
```diff
@@ -126,6 +131,10 @@ export default class IgcRatingComponent extends SizableMixin(
   @property({ type: Boolean, reflect: true })
   public readonly = false;
 
+  /** Toggles single selection visual mode. */
+  @property({ type: Boolean })
```
rkaraivanov, line 135: I suppose this should be a reflected property 
```suggestion
  @property({ type: Boolean, reflect: true })
```

rkaraivanov (changes requested):
LGTM with a couple of nitpicks

src/components/rating/rating.ts
simeonoff: This is not needed for styling, so no.

rkaraivanov (approved):

PR #434, opened 2022-07-14
fix(dropdown): Various dropdown fixes

### LINKED ISSUE #413: Dropdown: Add support for `Home` and `End` keyboard shorthands
## Issue
The dropdown menu doesn't support the going to the first or last items by pressing the `home` and `end` keys on the keyboard.

## Solution
- Pressing the `home` key on the keyboard should select the first item in the list of items.
- Pressing the `end` key on the keyboard should select the last item in the list of items.

### LINKED ISSUE #414: Dropdown: Duplicate selection
## Issue
Marking 2 items as selected, selects both items.

Example:
```html
<igc-dropdown>
    <igc-dropdown-item selected>Option 1</igc-dropdown-item>
    <igc-dropdown-item selected>Option 2</igc-dropdown-item>
</igc-dropdown>
```

## Expected Result

Only the last `selected` item should remain selected.

### LINKED ISSUE #419: Dropdown/Toggle: `igcOpened` and `igcClosed` emitted too early
## Description  
The `igcOpened` and `igcClosed` events are emitted before the dropdown container is shown(`igcOpened`) and hidden(`igcClosed`);

## Expected result  
The events should be emitted _after_ the container is shown/hidden.

rkaraivanov:
* Dropdown now supports navigation with Home/End keys.
* Dropdown now supports initial selection through `selected` attribute
  on its items. In case of multiple items, the last in the DOM hierarchy
  wins.
* Fixed an issue with `igcChange` being emitted twice when making selection.
* Fixed dropdown `opened/closed` events to be emitted after the underlying
  DOM is updated.
* Fixed dropdown closing events not being emitted when clicking around
  the page with `keepOpenOnOutsideClick` set to false.
* Fixed toggle directive `sameWidth` option to correctly reset the floater
  element styles after being turned off.

Closes #413
Closes #414
Closes #419


simeonoff (approved):

src/components/dropdown/dropdown.ts
```diff
@@ -178,115 +183,94 @@ export default class IgcDropdownComponent
...
-          this.onArrowUpKeyDown();
-          break;
-        case 'arrowdown':
-        case 'down':
-          this.onArrowDownKeyDown();
-          break;
-        default:
-          return;
-      }
+      this.keyDownHandlers.get(event.key)?.call(this);
```
damyanpetev, line 201: Neat!
Assuming the alternative names were for old browsers, not sure why we even had them in for WC stuff to begin with.

damyanpetev (commented):
PS: technically a feature + various fixes in the PR, if you want to align commits for the changelog.

src/components/dropdown/dropdown.ts
simeonoff: I believe the alternate names are actually down to OS architecture rather than browsers.

rkaraivanov: The virtual keycodes are down to OS architecture. The `KeyboardEvent.key` should be unified according to [this](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#navigation_keys)

rkaraivanov: Also
image

src/components/dropdown/dropdown-group.ts
```diff
@@ -26,7 +26,6 @@ export default class IgcDropdownGroupComponent extends LitElement {
   @queryAssignedElements({ flatten: true, selector: 'igc-dropdown-item' })
   public items!: Array<IgcDropdownItemComponent>;
 
-  /** @private */
```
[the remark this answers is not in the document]
rkaraivanov: Maybe as an alternative make the group items get the size from their parent instead of exposing an additional attribute.
Even with the **private** tag this would show in auto-complete in editors. Not to mention that the dropdown overwrites the value anyway...

src/components/dropdown/dropdown.ts
```diff
@@ -178,115 +183,94 @@ export default class IgcDropdownComponent
...
+    const item = this.allItems.filter((item) => item.selected).at(-1);
+    this.allItems.forEach((item) => (item.selected = false));
+    if (item) {
+      this.selectItem(item, false);
+    }
+  }
+
+  protected handleKeyDown = (event: KeyboardEvent) => {
+    const path = event.composedPath();
+    if (!(path.includes(this.target) || path.includes(this.content))) return;
```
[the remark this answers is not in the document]
rkaraivanov: https://github.com/IgniteUI/igniteui-webcomponents/blob/34c16542b3fe2bd15c56dd257b75dcc95a669c01/src/components/dropdown/dropdown.ts#L136-L143

The handler is attached and active only when it is in opened state

src/components/dropdown/dropdown-group.ts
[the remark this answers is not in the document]
damyanpetev: Agreed, the way you refactored it is likely better, however.. when the child looks for its parent explicitly it won't be super friendly for extended components (looking at the select, combo maybe). Perhaps generalize the `parent` with some base interface and move the query for it in a `protected getParent()` or something similar.

damyanpetev (approved):

PR #438, opened 2022-07-15
refactor: Unified autoincrement handling in components

rkaraivanov:


damyanpetev (approved):
Minor comment about unrelated code, otherwise LGTM.

PS: Still wondering what caused the excessive increments (Storybook somehow?). Also since the class can't be re-declared, wondering if we can get away with a static number prop similar to how the counter is ATM, rather then using a function scope.

damyanpetev (approved):

PR #445, opened 2022-07-20
Blazor 3.3.0 corrections

gmurray81:


src/components/dropdown/dropdown.ts
```diff
@@ -51,6 +52,9 @@ export interface IgcDropdownEventMap extends IgcToggleEventMap {
  * @csspart list - The dropdown list.
  */
 @themes({ bootstrap, fluent, indigo })
+@blazorAdditionalDependencies(
+  'IgcDropdownItemComponent, IgcDropdownGroupComponent'
```
damyanpetev, line 56: Header item?

src/components/common/decorators/blazorAdditionalDependencies.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Indicates additional dependencies that should be loaded by the module of a component, even if they don't appear to be dependencies.
+ * @param _additionalDependencies an alternate member name to use.
+ */
+export function blazorAdditionalDependencies(_additionalDependencies: string) {
```
damyanpetev, line 5: @gmurray81 Could try with some type of generic (or something that'll decompile away) here, so the deps are not stuck as hard strings. Doubtful they'll change, but still not ideal.

damyanpetev (approved):

src/components/common/decorators/blazorAdditionalDependencies.ts
damyanpetev: Actually, we might something different as well - the parent components already register the child dependencies - the issue is with metadata to to separate public children vs internally used components in templates, right?

So how about if we do a wrapper around the existing `defineComponents` calls (could be a decorator even if we're os inclined) that's sole purpose would be to provide meta on what the dependencies are. So something in the lines of:
```
[@]registerDependendecies(children: [TreeItem], internal/template: [Icon, CircularBar])
```
^ illustrative, but basically divide them by purpose in the call that will still just lead to `defineComponents(...children,...internal)`. That way we won't need the extra meta decorator to describe the dependency relations.

//CC @rkaraivanov

rkaraivanov: Seems like a good idea! Another approach is through some meta tag information in the doc strings of the component itself. Anyway, we can discuss it in a separate issue.

rkaraivanov (approved):

PR #447, opened 2022-07-21
Date Time Input - emit igcInput/igcChange correctly

### LINKED ISSUE #440: DateTime Input component: igcChange and IgcInput events not fired in all expected scenarios
## Description  

The `igc-date-time-input` does not emit the `igcInput` event when incrementing/decrementing the value with Arrow Up/Down or by scrolling. Also `igcChange` event is not emitted in these cases after commiting (i.e. the value has been in/decremented and the input has lost focus).
The `igcChange` event detail property, which should be the input's value is not correctly provided in case the input has been auto-completed. See below expected and current results.

## Steps to reproduce  

1. Add handlers for these events in date-time-input.stories.ts
```
<igc-date-time-input
  <!-- etc.-->
  @igcInput=${handleInput}
  @igcChange=${handleChange}
<!-- etc.-->
```

```
const handleInput = (ev:any) => {
  console.log(ev);
}
const handleChange = (ev:any) => {
  console.log(ev);
}
```
2. Run Date Time Input story and test the below listed expected scenarios.

## Result  

`igcInput`
- Does not fire when the user in/decrements the value using the Arrows or by scrolling

dti-indecrement-issue

`igcChange`
- Does not fire when the user has in/decremented the value using the Arrows or by scrolling after committing (i.e. the input losing focus). (see above gif as well)
- Does not provide the correct value, in case it has been “auto-completed”, i.e. the mask has been partially filled. Observe that in the below gif, even though the value is automatically filled to “01/01/2000”, the `igcChange` detail is null:

value-update

## Expected result  

`igcInput`
Expected to fire when receiving any user input, including:
- Character input when mask is filled
- Character input when mask is partially filled
- On de/incrementing using the Arrows or by scrolling

`igcChange`
- Expected to fire on committing a new value, i.e. there has been a change from the previous one.

onlyexeption:
Closes #440

### DIFF
--- a/src/components/date-time-input/date-time-input.ts
+++ b/src/components/date-time-input/date-time-input.ts
@@ -53,6 +53,7 @@ export default class IgcDateTimeInputComponent extends EventEmitterMixin<
   protected _defaultMask!: string;
   protected _value!: Date | null;
 
+  private _oldValue: Date | null = null;
   private _inputDateParts!: DatePartInfo[];
   private _inputFormat!: string;
   private _datePartDeltas: DatePartDeltas = {
@@ -322,6 +323,7 @@ export default class IgcDateTimeInputComponent extends EventEmitterMixin<
 
     const newValue = this.trySpinValue(targetPart, delta);
     this.value = newValue;
+    this.handleInput();
   }
 
   /** Decrements a date/time portion. */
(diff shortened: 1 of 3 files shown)


src/components/date-time-input/date-time-input.ts
```diff
@@ -617,6 +625,10 @@ export default class IgcDateTimeInputComponent extends EventEmitterMixin<
       this.updateMask();
     }
 
+    if (this._oldValue !== this.value) {
+      this.handleChange();
+    }
```
rkaraivanov, line 630: I reckon part of the logic in the blur handler can be refactored into the "default" change handler, specifically the partially completed mask handling.
Anyway, I'm ok with the current implementation and this can be addressed in a separate PR since it would be an internal change

rkaraivanov (approved):

PR #453, opened 2022-07-27
refactor(slider): update theme after latest fluent spec

SisIvanova:


simeonoff (approved):

simeonoff (changes requested):
The focus styles are not removed on blur in the Range Slider component.

simeonoff (approved):

simeonoff (approved):

Issue #378, opened 2022-05-19
Linear progress - indeterminate animation is broken in safari


desig9stein:
## Steps to reproduce  
Open linear progress sample in safari

## Result 
The indeterminate animation is different from chrome and looks a bit broken.

## Expected result  
The animation should be the same as in chrome

PR #465, opened 2022-08-05
fix linear progress animations

desig9stein:
Closes: #378


didimmova:
Animations in Safari still don't work properly in every theme except for fluent. When the window is thinner there's no fill at all. Chrome and Firefox are fine.

simeonoff (approved):

PR #466, opened 2022-08-08
blazor annotations

gmurray81:
annotations necessary to correct the blazor build

### DIFF
--- a/src/components/accordion/accordion.ts
+++ b/src/components/accordion/accordion.ts
@@ -4,6 +4,7 @@ import { styles } from './themes/accordion.base.css.js';
 
 import { defineComponents } from '../common/definitions/defineComponents.js';
 import IgcExpansionPanelComponent from '../expansion-panel/expansion-panel.js';
+import { blazorSuppress } from '../common/decorators/blazorSuppress.js';
 
 defineComponents(IgcExpansionPanelComponent);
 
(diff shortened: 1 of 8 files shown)


damyanpetev (approved):
I don't quite get the need for wrappers to import base classes, but if this is needed with the current state - sure we can merge this.

The moment you hit deep imports though, you're basically using API that's considered private (as you noted in the description - not exported from the root public API). Some lint rules even forbid such imports.
Anyway, if we can avoid exposing these in the wrappers or there's a valid reason for those to be publicly available we should probably discuss it.

gmurray81:
I think if there's a reason for the base classes to exist at the ts level there's probably reason for the hierarchy to exist at the Blazor/React level? Maybe the base classes are overwrought at the WC level too if we felt the need to hide them?

We could add an attribute to flatten some of the base classes alternatively, but that was more work than deep importing the base, and I'm focusing on all the prototyping we need done now and fully automating the Blazor build, so I need everything to be in a state where it is generating valid output to check in the build changes.

damyanpetev:
The base classes are mostly for internal code organization and reuse, so they have a reason to be alright. Could replace them with mixins, though I doubt that'll help :P Though we still ight end up there eventually.
There's probably a case to be made for those to be exposed on the off chance someone wants to extend our components and the base fits their needs better, but we're not really focusing on that right now too.

As for the meta on when to flatten or if something is exported at root, in theory that information can be gained from looking at the public API exports even, but again - not the faster option right now

rkaraivanov (approved):

PR #471, opened 2022-08-16
refactor: Reduced LTR query calls and improved rating rendering

rkaraivanov:
- Extracted `isLTR` to an external helper function.
- Reduced number of LTR query calls in certain scenarios to avoid
  reflow thrashing.
- Cached rating symbols rendering logic. Properties not related
  to the symbols state will not trigger a DOM re-render.

### DIFF
--- a/src/components/common/util.ts
+++ b/src/components/common/util.ts
@@ -48,3 +48,7 @@ export function createCounter() {
     return i;
   };
 }
+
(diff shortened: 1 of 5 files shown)


simeonoff:
@rkaraivanov hover preview with single selection doesn't seem to work as intended in this PR. The previously selected symbol stays colored even after changing the selection.

PR #488, opened 2022-08-23
feat(*): role property to icon, focusable remove chip

### LINKED ISSUE #480: Chip Remove icon
## Description  
The remove icon of the chip can't be reached with a keyboard. 

## Expected result  
The remove icon of the chip should be reachable with a keyboard.

ChronosSF:
Closes #480

### DIFF
--- a/src/components/chip/chip.ts
+++ b/src/components/chip/chip.ts
@@ -78,14 +78,14 @@ export default class IgcChipComponent extends SizableMixin(
   }
 
   protected handleRemove(e: Event) {
-    if (
-      e.type === 'keyup' &&
-      ![' ', 'enter'].includes((e as KeyboardEvent).key.toLowerCase())
-    ) {
-      return;
-    }
     this.emitEvent('igcRemove');
     e.stopPropagation();
+  }
+
+  protected handleKeyup(e: KeyboardEvent) {
+    if (/\s|enter/i.test(e.key)) {
+      this.handleRemove(e);
+    }
   }
 
   protected override render() {
@@ -118,7 +118,7 @@ export default class IgcChipComponent extends SizableMixin(
             ? html`<slot
                 @slotchange=${this.slotChanges}
                 @click=${this.handleRemove}
-                @keyup=${this.handleRemove}
+                @keyup=${this.handleKeyup}
                 name="remove"
               >
                 <igc-icon
--- a/src/components/chip/chip.ts
+++ b/src/components/chip/chip.ts
@@ -118,6 +118,9 @@ export default class IgcChipComponent extends SizableMixin(
                   size=${this.size}
                   name="chip_cancel"
                   collection="internal"
+                  tabindex="0"
+                  role="button"
+                  aria-label="remove"
                 ></igc-icon>
               </slot>`
             : nothing}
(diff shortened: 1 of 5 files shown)


src/components/icon/icon.ts
```diff
@@ -71,14 +71,22 @@ export default class IgcIconComponent extends SizableMixin(LitElement) {
...
+  public role = 'img';
+
   constructor() {
     super();
     this.size = 'medium';
   }
 
   public override connectedCallback() {
     super.connectedCallback();
-    this.setAttribute('role', 'img');
```
rkaraivanov, line 81: Can we check for an existing role attribute and then set the default if none is found?

Something about a public property used in an internal context that is reflected as a standard attribute feels off to me.

rkaraivanov (changes requested):

rkaraivanov:
@ChronosSF
 
While this PR fixes the focus issue, triggering an "activation" event through <kbd>Space</kbd> | <kbd>Enter</kbd> on the remove button does not fire the **igcRemove** event.
It does it on click, so this also falls under general accessibility. 

I suggest we fix that too either in this PR or in a new one.

src/components/chip/chip.ts
```diff
@@ -112,12 +118,16 @@ export default class IgcChipComponent extends SizableMixin(
             ? html`<slot
                 @slotchange=${this.slotChanges}
                 @click=${this.handleRemove}
+                @keyup=${this.handleRemove}
```
rkaraivanov, line 121:
```suggestion
                @keyup=${this.handleKeyup}
```

src/components/icon/icon.ts
```diff
@@ -78,7 +78,9 @@ export default class IgcIconComponent extends SizableMixin(LitElement) {
...
+    if (!this.getAttribute('role')) {
```
rkaraivanov, line 81:
```suggestion
    if (!this.hasAttribute('role')) {
```

stories/icon.stories.ts
```diff
@@ -33,6 +33,12 @@ const metadata = {
       control: 'boolean',
       defaultValue: false,
     },
+    role: {
```
rkaraivanov, line 36: Can you regenerate the icon stories and commit them as well?

src/components/chip/chip.ts
```diff
@@ -78,6 +78,12 @@ export default class IgcChipComponent extends SizableMixin(
...
   protected handleRemove(e: Event) {
+    if (
+      e.type === 'keyup' &&
+      ![' ', 'enter'].includes((e as KeyboardEvent).key.toLowerCase())
+    ) {
+      return;
+    }
     this.emitEvent('igcRemove');
     e.stopPropagation();
   }
```
rkaraivanov, line 89:
```suggestion
  protected handleRemove(e: Event) {
    this.emitEvent('igcRemove');
    e.stopPropagation();
  }
  
  protected handleKeyup(e: KeyboardEvent) {
    if (/\s|enter/i.test(e.key)) {
      this.handleRemove(e);
    }
  }
```
A bit cleaner IMO

rkaraivanov (changes requested):

src/components/chip/chip.ts
ChronosSF: Thought regex is evil ^^

rkaraivanov: It is a tool. It can be used for both good and evil 😄

rkaraivanov (approved):

Issue #482, opened 2022-08-16
Switch accessibility improvements


desig9stein:
## Description  
The reader reads the switch like it is a checkbox since the input inside is of type checkbox.

The input should be of `type="button"` and  `role="switch"` then the reader will read correctly.

ChronosSF:
@desig9stein , from my tests, NVDA reads both the current switch and the proposed change in the same way. Windows Narrator doesn't read the value change in either case, so there doesn't seem to be any improvement we can make.

desig9stein:
@ChronosSF You are correct that NVDA does read it the same way, however, the screen reader on mac will actually say "switch". Since the current implementation makes sense and my suggestion is more of an enhancement and it will affect MAC users only, I am not sure if is worth it.

PR #489, opened 2022-08-23
fix(switch): switching to type button

ChronosSF:
Closing #482


ChronosSF:
After discussing this with Simeon we decided not to change this behavior.

PR #490, opened 2022-08-25
fix(list): add aria-label attribute

### LINKED ISSUE #472: List аccessibility improvements
## Description  
The screen reader tells you the number of list items but not which item you are on, also it does not read the text inside the item.

onlyexeption:
Closes #472


simeonoff (approved):

rkaraivanov:
Fixed in #980

PR #492, opened 2022-08-25
fix(toast): adding alert role for toast

### LINKED ISSUE #479: Toast accessibility improvements
## Description  
The screen reader does not read the text inside the toast on opening.

ChronosSF:
Closes #479

### DIFF
--- a/src/components/toast/toast.ts
+++ b/src/components/toast/toast.ts
@@ -36,13 +36,6 @@ export default class IgcToastComponent extends LitElement {
   @property({ type: Boolean, reflect: true, attribute: 'keep-open' })
   public keepOpen = false;
 
-  constructor() {
-    super();
-    if (!this.hasAttribute('role')) {
-      this.setAttribute('role', 'alert');
-    }
-  }
-
   /** Closes the toast. */
   public hide() {
     if (this.open) {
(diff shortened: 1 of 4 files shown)


src/components/toast/toast.ts
```diff
@@ -36,6 +36,13 @@ export default class IgcToastComponent extends LitElement {
   @property({ type: Boolean, reflect: true, attribute: 'keep-open' })
   public keepOpen = false;
 
+  constructor() {
+    super();
+    if (!this.hasAttribute('role')) {
+      this.setAttribute('role', 'alert');
+    }
+  }
```
rkaraivanov, line 44: You should move this to the connectedCallback hook.
As per the [specification](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-element-conformance):

> The element must not gain any attributes or children, as this violates the expectations of consumers who use the [createElement](https://dom.spec.whatwg.org/#dom-document-createelement) or [createElementNS](https://dom.spec.whatwg.org/#dom-document-createelementns) methods.

rkaraivanov (changes requested):

desig9stein:
@ChronosSF The reader on mac does not read anything with just the role, however, adding `aria-live="polite"` seems to fix the problem. 

We use the same approach in Angular.

[igx toast](https://www.infragistics.com/products/ignite-ui-angular/angular/components/toast#show-toast)
[aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live)

ChronosSF:
Sadly, providing any aria-live seems to completely kill off the read out from Windows Narrator, but that seems to be the case with angular too. I guess having it read on Mac is better as we kinda go in the way of using nvda as baseline.

rkaraivanov (approved):

PR #493, opened 2022-08-26
fix(range-slider): add aria-valuetext to range slider

### LINKED ISSUE #477: Range slider accessibility improvements
## Description  
The screen reader does not tell you on which thumb you are it only reads the value of the thumb.

onlyexeption:
Closes #477


onlyexeption:
Closed due to branching from wrong branch, not master

PR #494, opened 2022-08-26
fix(range-slider): add ariaValueText

### LINKED ISSUE #477: Range slider accessibility improvements

onlyexeption:
Closes #477

### DIFF
--- a/src/components/slider/slider-base.ts
+++ b/src/components/slider/slider-base.ts
@@ -544,8 +544,18 @@ export class IgcSliderBaseComponent extends LitElement {
     return groups;
   }
 
-  protected renderThumb(value: number, ariaLabel?: string, thumbId?: string) {
+  protected renderThumb(
+    value: number,
+    ariaLabel?: string,
+    thumbId?: string,
+    ariaValueText?: string
+  ) {
     const percent = `${this.valueToFraction(value) * 100}%`;
+    const textValue = this.labels
+      ? this.labels[value]
+      : this.valueFormat || this.valueFormatOptions
+      ? this.formatValue(value)
+      : ariaValueText;
 
     return html`
       <div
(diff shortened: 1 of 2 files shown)


src/components/slider/slider-base.ts
```diff
@@ -544,8 +544,37 @@ export class IgcSliderBaseComponent extends LitElement {
...
+
+    const thumbs = this.shadowRoot?.querySelectorAll('div[part="thumb"]');
+    thumbs?.forEach((t) => {
+      if (t.id !== thumbId) {
+        this.activeThumb?.setAttribute(
+          'aria-valuetext',
+          `${Math.min(
+            parseInt(thumbVal!),
+            parseInt(t.ariaValueNow!)
+          )} - ${Math.max(parseInt(thumbVal!), parseInt(t.ariaValueNow!))}`
```
rkaraivanov, line 560: Since step can be decimal, maybe parse to float and format it to step precision.
This should account for the **discrete** property as well, since when active the value is unbound from the step property.

src/components/slider/slider-base.ts
```diff
@@ -544,8 +544,37 @@ export class IgcSliderBaseComponent extends LitElement {
...
+        );
+      }
+    });
+  }
+
+  protected renderThumb(
+    value: number,
+    ariaLabel?: string,
+    thumbId?: string,
+    ariaValueText?: string
```
rkaraivanov, line 570: A bit of nitpick but since all params are kind of internal to the component can you simplify the call signature to just the thumbId and get rest based on that?

rkaraivanov (changes requested):

src/components/slider/slider-base.ts
onlyexeption: I could, but how would you pass the aria label and the valueText like we do in `renderThumbs()` in range-slider.ts?

rkaraivanov (approved):

PR #495, opened 2022-08-26
refactor(select): keyboard behavior matches angular

simeonoff:


rkaraivanov:
@simeonoff 
LGTM, however when the dropdown is opened, the left/right keys don't seem to mimic the up/down keys correctly. For example, it cannot skip over disabled items.

simeonoff:
> @simeonoff
> LGTM, however when the dropdown is opened, the left/right keys don't seem to mimic the up/down keys correctly. For example, it cannot skip over disabled items.

Should work fine now. Had to update the dropdown keyboard navigation as well to enable support for navigation with left and right arrow keys.

rkaraivanov (approved):

PR #497, opened 2022-08-30
fix(calendar): Range selection a11y improvement

### LINKED ISSUE #476: Calendar accessibility improvements
## Description  
The screen reader does not read the range selection start and end dates and the current date.

rkaraivanov:
Closes #476

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Chip - made remove button accessible with the keyboard [#480](https://github.com/IgniteUI/igniteui-webcomponents/issues/480)
+- Calendar - range selection a11y improvements [#476](https://github.com/IgniteUI/igniteui-webcomponents/issues/476)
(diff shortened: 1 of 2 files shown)


ChronosSF:
nvda reads selection really well when the range is selected but while the user is choosing the range with the keyboard it reads the first selected date while the second is presented just by its position and weekday. Something like:

Selected Aug 18, 2022 . Not selected column 5 wednesday

Can we have the readout work while the user is choosing what to select as the second date in the range? Basically repeating the select behavior but leaving "unselected" so that he knows he hasn't yet confirmed the selection by just navigating with the arrow keys.

ChronosSF:
Sounds good now, even with the windows narrator.

@desig9stein , will you be able to test it with the Mac narrator?

damyanpetev (approved):

desig9stein:
I can hear the mac narrator reading the range, however, I can't hear anything about the current date. In angular calendar when I go over the current date the narrator reads something like "23 September 2022 **Current date**"

ChronosSF:
@desig9stein , can we merge it like this and have a separate issue so that we don't block the release?

PR #509, opened 2022-09-14
fix(tabs): resolving an issue with selected attr in tab

### LINKED ISSUE #507: Tab is showing all tab's content on load
## Description 

The content of all tabs are shown by default.

## Steps to reproduce  

1. See the following demo: https://codesandbox.io/s/hardcore-platform-dgmwfb?file=/index.html

## Result  
The content of all tabs are shown by default. When tabs are clicked everything seems ok.

All `igc-tab-panel`s that needs to be hidden is missing the `display: none` style on load.

## Expected result  
Show only the content for the selected tab

ChronosSF:
Closes #507

### DIFF
--- a/src/components/tabs/tabs.ts
+++ b/src/components/tabs/tabs.ts
@@ -142,6 +142,8 @@ export default class IgcTabsComponent extends EventEmitterMixin<
 
     await this.updateComplete;
 
+    this.activeTab = this.tabs.find((tab) => tab.selected);
+    this.updateSelectedTab();
     this.syncAttributes();
     this.setupObserver();
     this.setSelectedTab(
--- a/src/components/tabs/tabs.spec.ts
+++ b/src/components/tabs/tabs.spec.ts
@@ -89,6 +89,21 @@ describe('Tabs component', () => {
       verifySelection(element, 'second');
     });
 
+    it('selects the tab with selected attribute in the template', async () => {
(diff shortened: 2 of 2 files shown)


src/components/tabs/tabs.ts
```diff
@@ -142,6 +142,8 @@ export default class IgcTabsComponent extends EventEmitterMixin<
 
     await this.updateComplete;
 
+    this.activeTab = this.tabs.find((tab) => tab.selected);
```
damyanpetev, line 145: `find` will take the first match, but in the case of multiple elements marked as selected we've already established the last one wins. Then there's the matter that just a few lines below there's a `this.setSelectedTab` call that correctly picks the last selected tab or falls back the first enabled, so assigning the active tab here is redundant.

src/components/tabs/tabs.ts
```diff
@@ -142,6 +142,8 @@ export default class IgcTabsComponent extends EventEmitterMixin<
...
+    this.updateSelectedTab();
```
damyanpetev, line 146: It seems like you can just move this call after the `setSelectedTab` so it'll properly assign the visibility of the panels

damyanpetev: Could also use a name change it seems, since the name of the method doesn't really correlate with what it's doing (fixing up the panels)

damyanpetev (changes requested):

damyanpetev (approved):

PR #534, opened 2022-10-27
fix(tabs): panel inline styles

### LINKED ISSUE #528: Tabs panel have inline style added
Looking at tabs, there is this display: block inline style added:

image

Let's say I want to layout the elements inside the tabs panel in flex, it makes it impossible:

image

SisIvanova:
Closes #528

### DIFF
--- a/src/components/tabs/tabs.ts
+++ b/src/components/tabs/tabs.ts
@@ -236,11 +236,9 @@ export default class IgcTabsComponent extends EventEmitterMixin<
 
   protected updateSelectedTab() {
     this.tabs.forEach((tab) => (tab.selected = tab === this.activeTab));
-    this.panels.forEach((panel) =>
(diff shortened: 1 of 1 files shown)


src/components/tabs/tabs.spec.ts
```diff
@@ -31,7 +31,7 @@ describe('Tabs component', () => {
 
   const getSelectedPanel = (tabs: IgcTabsComponent) => {
     const collection = getPanels(tabs).filter(
-      (panel) => panel.style.display === 'block'
+      (panel) => panel.hasAttribute('hidden') == false
```
simeonoff, line 34:
```suggestion
      (panel) => panel.hasAttribute('hidden') === false
```

simeonoff (approved):

PR #553, opened 2022-11-18
fix(checkbox/radio/switch): extra margin without label

### LINKED ISSUE #549: Checkbox, radio, and switch controls have label margin even when there is no label
## Description  
There is a space in the form of a margin between a label and a control.
The problem is that the space is there even when we don't have a label, this is because even if there is no label the markup for it stays in the dom.

If you are wandering for a scenario in which this could be a problem here is one. In the Apex grid, we have a column with checkboxes and those checkboxes have their labels as the column header, they don't need a label next to each checkbox.

Although this is currently observable only in the fluent theme, only when we focus the control with the keyboard I believe it should be fixed as it could lead to other layout problems. For example, see how the checkboxes in the grid are not aligned with the header text as the other components inside the cells.

## Steps to reproduce  

1. Open any of the above controls
2. Remove the label

## Result  
The control preserves the spacing that separates it from the label

## Expected result  
The control should not preserve the spacing that separates it from the label.

## Attachments  
The screenshots show what is happening if you don't have a label and focus on the switch using the keyboard. Currently, this is observable only inside the fluent theme since by design the keyboard focus styles wrap the control and the label.

Screenshot 2022-11-14 at 16 24 31

Screenshot 2022-11-14 at 16 24 37

SisIvanova:
Closes #549

### DIFF
--- a/src/components/checkbox/checkbox-base.ts
+++ b/src/components/checkbox/checkbox-base.ts
@@ -1,5 +1,5 @@
 import { LitElement } from 'lit';
-import { property, query, state } from 'lit/decorators.js';
+import { property, query, queryAssignedNodes, state } from 'lit/decorators.js';
 import { alternateName } from '../common/decorators/alternateName.js';
 import { blazorDeepImport } from '../common/decorators/blazorDeepImport.js';
 import { blazorTwoWayBind } from '../common/decorators/blazorTwoWayBind.js';
@@ -20,8 +20,14 @@ export class IgcCheckboxBaseComponent extends EventEmitterMixin<
   @query('input[type="checkbox"]', true)
   protected input!: HTMLInputElement;
 
+  @queryAssignedNodes({ flatten: true })
+  protected label!: Array<Node>;
+
   @state()
   protected focused = false;
+
+  @state()
+  protected hide = false;
 
   /** The name attribute of the control. */
   @property()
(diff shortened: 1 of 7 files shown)


src/components/checkbox/checkbox-base.ts
```diff
@@ -121,4 +127,12 @@ export class IgcCheckboxBaseComponent extends EventEmitterMixin<
       this.focused = true;
     }
   }
+
+  protected override async firstUpdated() {
+    if (this.label.length == 0) {
```
simeonoff, line 132:
```suggestion
    if (this.label.length === 0) {
```

src/components/radio/radio.ts
```diff
@@ -189,6 +195,14 @@ export default class IgcRadioComponent extends EventEmitterMixin<
     return Array.from<IgcRadioComponent>(group.querySelectorAll('igc-radio'));
   }
 
+  protected override async firstUpdated() {
+    if (this.label.length == 0) {
```
simeonoff, line 199:
```suggestion
    if (this.label.length === 0) {
```

simeonoff (changes requested):
Although I like the brevity of the property being named just `hide`, it's not very descriptive as to what this state is being used for. Perhaps `hideLabel` is more befitting for this occasion.

simeonoff (approved):

PR #567, opened 2022-11-29
build: Updated to TS 4.9 and build chain

rkaraivanov:
There seems to be some issue with the Lit linting tools with the latest TS version.
That's why there is bunch of `@attr` tags for public properties that are also attributes.


src/components/icon/icon.ts
```diff
@@ -43,6 +41,8 @@ export default class IgcIconComponent extends SizableMixin(LitElement) {
    *
    * @attr [name=""]
    */
+  @property()
+  @alternateName('iconName')
```
damyanpetev, line 45: @rkaraivanov This seems to mess with the current function of `@alternateName`, seems to be read off of setters only.
// CC @gmurray81

damyanpetev: Actually, `  /* alternateName: iconName */` on the setter seems to also do the job, so we might default to that

damyanpetev: Simple prop works too :)

damyanpetev (approved):

PR #570, opened 2022-11-30
refactor(stepper): full types for event args, drop alt name comment

damyanpetev:
Don't use anonymous shape types for public API in general - kind of messes with the generation for other plats at least atm :)
PS: I don't see any types for the events being exposed anywhere btw, perhaps these should be?
// CC @rkaraivanov


src/components/stepper/stepper.common.ts
```diff
@@ -1,14 +1,17 @@
...
+export interface IgcActiveStepChangingArgs {
+  oldIndex: number;
+  newIndex: number;
+  owner: IgcStepperComponent;
+}
+
+export interface IgcActiveStepChangedArgs {
+  index: number;
+  owner: IgcStepperComponent;
+}
```
damyanpetev, line 12:
```suggestion
export interface IgcActiveStepChangedArgs {
  index: number;
}
```

src/components/stepper/stepper.common.ts
```diff
@@ -1,14 +1,17 @@
 import type IgcStepperComponent from './stepper.js';
 
+export interface IgcActiveStepChangingArgs {
+  oldIndex: number;
+  newIndex: number;
+  owner: IgcStepperComponent;
+}
```
damyanpetev, line 7:
```suggestion
export interface IgcActiveStepChangingArgs {
  oldIndex: number;
  newIndex: number;
}
```

rkaraivanov (approved):

viktorkombov (approved):

PR #579, opened 2022-12-06
ci(a11y): enable lit-a11y eslint plugin

damyanpetev:
Revival of #452

### DIFF
--- a/src/components/nav-drawer/nav-drawer.ts
+++ b/src/components/nav-drawer/nav-drawer.ts
@@ -80,6 +80,7 @@ export default class IgcNavDrawerComponent extends LitElement {
     };
   }
 
+  /* eslint-disable lit-a11y/click-events-have-key-events */
   protected override render() {
     return html`
       <div part="overlay" @click=${this.hide}></div>
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -411,7 +411,8 @@ export default class IgcSelectComponent extends EventEmitterMixin<
       <div
         role="combobox"
         tabindex=${this.disabled ? -1 : 0}
-        aria-owns="dropdown"
+        aria-controls="dropdown"
+        aria-expanded=${this.open}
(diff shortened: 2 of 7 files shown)


src/components/nav-drawer/nav-drawer.ts
```diff
@@ -80,6 +80,7 @@ export default class IgcNavDrawerComponent extends LitElement {
     };
   }
 
+  /* eslint-disable lit-a11y/click-events-have-key-events */
```
damyanpetev, line 83: Might be a future addition to handle ESC to close the drawer for example. https://m3.material.io/components/navigation-drawer/accessibility is still pending, but I've seen some apps close with KB nav. However, there's a good chance this is left as an app scenario as we prob can't manage the focus state after close correctly, in which case this could be fine like that.

src/components/select/select.ts
```diff
@@ -465,6 +466,7 @@ export default class IgcSelectComponent extends EventEmitterMixin<
         part="base"
         style=${styleMap({ position: this.positionStrategy })}
         @click=${this.handleClick}
+        aria-hidden="true"
```
damyanpetev, line 469: Kinda impossible to add /* eslint-disable lit-a11y/click-events-have-key-events */ here without dropping the entire template which kind of defeats the purpose of the checks, so went with `aria-hidden` which silences such rules. 
Not sure if I prefer adding extra aria attr where it shouldn't be needed in theory or having to refactor around eslint comments so only such cases (delegated click for other elements) can be handled.
// CC @rkaraivanov

rkaraivanov: At least for now, I think we can stick with the extra attributes.
Either `aria-hidden` or `role="presentation"` based on the context and DOM

rkaraivanov:
Closing this PR since there is already a Storybook addon for accessibility smoke tests and
the test suite runs such tests.

PR #585, opened 2022-12-08
fix(select): make checkValidity to only return whether the component is valid

### LINKED ISSUE #543: Select component - The checkValidity method moves the focus to the control
## Description  
When the `checkValidity` method is called and the select control isn't satisfies its constraints the focus is moved to it.

## Steps to reproduce  

1. Open the following sample - [here](https://codesandbox.io/s/igc-select-checkvalidity-fuo06d?file=/src/index.ts).
2. Type something in the input field.

## Result  
The focus will be moved to the select on the first keystroke.

## Expected result  
The focus shouldn't be moved to the select when the `checkValidity` method is called.

viktorkombov:
Closes #543

### DIFF
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -232,16 +232,17 @@ export default class IgcSelectComponent extends EventEmitterMixin<
     this.target.blur();
   }
 
-  /** Checks the validity of the control. */
+  /** Checks the validity of the control and moves the focus to it if it is not valid. */
   public reportValidity() {
     this.invalid = this.required && !this.value;
     if (this.invalid) this.target.focus();
     return !this.invalid;
   }
 
-  /** A wrapper around the custom reportValidity method to comply with the native select API. */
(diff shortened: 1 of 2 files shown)


src/components/select/select.ts
```diff
@@ -232,16 +232,17 @@ export default class IgcSelectComponent extends EventEmitterMixin<
     this.target.blur();
   }
 
-  /** Checks the validity of the control. */
+  /** Checks the validity of the control and moves the focus to it if it is not valid. */
   public reportValidity() {
     this.invalid = this.required && !this.value;
     if (this.invalid) this.target.focus();
     return !this.invalid;
```
simeonoff, line 239:
```suggestion
    const valid = this.checkValidity();
    if (!valid) this.target.focus();
    return valid;
```

simeonoff (changes requested):

simeonoff (approved):

PR #594, opened 2022-12-20
fix(input): material themes don't match design

simeonoff:


src/components/input/themes/light/input.material.scss
```diff
@@ -1,10 +1,10 @@
 @use 'sass:map';
 @use '../../../../styles/utilities' as *;
 
-$idle-color: color(gray, 600) !default;
+$idle-color: color(gray, 500) !default;
 $idle-hover-color: color(gray, 800) !default;
 $hover-background: color(gray, 200) !default;
-$focus-background: var(--focus-background, color(gray, 300)) !default;
+$focus-background: var(--focus-background, color(gray, 200)) !default;
```
SisIvanova, line 7: I think this change is unnecessary

-Box input:
Focus state:
-background (gray-300)

simeonoff: Reverted back to the original value.

SisIvanova (approved):

desig9stein (approved):

PR #602, opened 2023-01-05
refactor: Side-effect free component registration

rkaraivanov:


src/components/accordion/accordion.spec.ts
```diff
@@ -5,12 +5,11 @@ import {
   unsafeStatic,
   elementUpdated,
 } from '@open-wc/testing';
-import { defineComponents, IgcExpansionPanelComponent } from '../../index.js';
 import IgcAccordionComponent from './accordion.js';
 
 describe('Accordion', () => {
   before(() => {
-    defineComponents(IgcAccordionComponent, IgcExpansionPanelComponent);
+    IgcAccordionComponent.register();
```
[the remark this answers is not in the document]
damyanpetev: To that end, is the `register` supposed to be public? Trying to figure out if it'll make the guidance somewhat confusing.

src/components/common/definitions/defineComponents.ts
```diff
@@ -1,9 +1,7 @@
-export const defineComponents = (...components: CustomElementConstructor[]) => {
-  for (const component of components) {
-    const tagName = (component as any).tagName;
+import type { IgniteComponent } from '../types';
 
-    if (tagName && !window.customElements.get(tagName)) {
-      window.customElements.define(tagName, component);
-    }
+export const defineComponents = (...components: IgniteComponent[]) => {
```
damyanpetev, line 3: So incoming types match `IgniteComponent` by schema alone?
I actually don't mind, spares us the need explicitly implement for each component

src/components/common/types.ts
```diff
@@ -1 +1,6 @@
 export type Direction = 'ltr' | 'rtl' | 'auto';
+
+export type IgniteComponent = CustomElementConstructor & {
```
damyanpetev, line 3: Is this making it to the root export? Can't quite see it

damyanpetev (commented):
Overall looking good - a few questions really, but will do what it says on the tin and remove the side effects. 
Additional question: Should we pre-emptively `await` register calls and/or make them async? Just so we can have it in place? I remember you mentioned you wanted to do dynamic imports for deps at some point

src/components/common/types.ts
rkaraivanov: Sure, I will export it

src/components/common/definitions/defineComponents.ts
rkaraivanov: Yep, it is a schema-based typing

src/components/accordion/accordion.spec.ts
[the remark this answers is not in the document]
rkaraivanov: `register` is public but the official guidance is still to use `defineComponents`.

We won't document it and if the end-user calls it, the effect would be the same as calling
`defineComponents(x)` twice.

rkaraivanov:
> Additional question: Should we pre-emptively `await` register calls and/or make them async? Just so we can have it in place? I remember you mentioned you wanted to do dynamic imports for deps at some point

IMO leaving as it is right now looks good enough. Not sure how much performance benefit dynamic imports will bring while it will definitely complicate the implementation.

rkaraivanov:
Closed in favor of #895

PR #612, opened 2023-01-10
feat(animations): add a basic animation player

### LINKED ISSUE #599: [Animations] - Expansion Panel
Animate the expanding/collapsing states of the expansion panel.

simeonoff:
Closes #599 

Add a basic animation player that plays Web Animations and calls a function when the animation finishes playing. As a bonus feature, the animation player will animate height automatically if its value is set to 'auto'.

This PR animates the expansion panel as a showcase for how the player can be used.


rkaraivanov (changes requested):
LGTM

Some minor suggestions

rkaraivanov (approved):

PR #615, opened 2023-01-12
fix(calendar): navigate to the date set as value initially

### LINKED ISSUE #436: Calendar does not navigate to date set as value
## Update
Calendar should follow the specification, as described [here](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Calendar-Specification#active-date-with-selection), when deciding on what active data should point to.

## Description  
When setting a value to some date I expect calendar to navigate to this date without explicitly set `activeDate`.
There is functional difference from what happens in web components calendar and angular calendar when date is set:

Web components calendar with value set
image

Angular calendar with value set
image

## Steps to reproduce  

1. Create a calendar and set its value

## Result  
Calendar shows current month and year.

## Expected result  
Calendar should navigate to the date as specified in its `value`.

## Note
Calendar without `value` set is also different in Angular and Web Components. Having no value Angular shows today date in its header while web components calendar shows text - 'Selected date'.

ddaribo:
Closes #436


src/components/calendar/common/calendar-base.ts
```diff
@@ -105,4 +105,12 @@ export class IgcCalendarBaseComponent extends LitElement {
   protected selectionChange() {
     this.value = undefined;
   }
+
+  protected override async firstUpdated() {
+    if (this.selection === 'single') {
+      this.activeDate = this.value ? this.value : new Date();
```
wnvko, line 111: you can simplify this with [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

wnvko (approved):

wnvko (approved):

rkaraivanov:
@ddaribo 
[Here](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Calendar-Specification#active-date-with-selection) is the updated behavior of how active date and set values should interact in the calendar.

wnvko (commented):
Issue updated as per the new specification.
Please update component to follow the spec and add tests for each active date related scenario described in the tests.

wnvko (changes requested):
Issue updated as per the new specification.
Please update component to follow the spec and add tests for each active date related scenario described in the spec.

wnvko (approved):

Issue #610, opened 2023-01-10
Implement Animations


SisIvanova:
Animate the states of the following components: 

- [x] Dialog
- [ ] Drop Down
- [ ] NavDrawer
- [x] Select
- [x] Snackbar
- [x] Stepper
- [x] Toast
- [x] Tree

PR #625, opened 2023-01-18
feat(dialog/snackbar/toast): animate component states

SisIvanova:
Related to #610


simeonoff (commented):
Toast and snackbar look good. I feel, though, the dialog needs some more attention. The opening animation is not smooth, especially around the backdrop element.

simeonoff:
@SisIvanova The dialog's animation is choppy when closing the dialog window. The backdrop background disappears abruptly and is not synced with the dialog window.

src/components/dialog/dialog.ts
```diff
@@ -116,27 +121,43 @@ export default class IgcDialogComponent extends EventEmitterMixin<
...
+    return event.type === 'finish';
+  }
+
   /** Opens the dialog. */
   public show() {
     if (this.open) {
       return;
     }
 
+    this.backdrop.setAttribute('aria-hidden', 'false');
```
simeonoff, line 148: Why do we need to have this here and elsewhere, when it's already being set in the template by binding?

simeonoff (changes requested):

PR #627, opened 2023-01-19
Refactor Tab Component

### LINKED ISSUE #504: Tab component - Unify implementation in a single component
## Description  
When the tabs were implemented, due to the way slot projection work, we took the approach to have separate header and panel components for each tab.

However, this creates problems for accessibility and ease of use from a developer point of view.

I suggest refactoring the tab into a single `<igc-tab>` component with the appropriate slots for header/content.

mmart1n:
Closes #504


rkaraivanov:
Closed in favor of #1352

PR #634, opened 2023-01-25
refactor(theme): normalize input themes

### LINKED ISSUE #619: Input helper text inconsistency
## Description  
The input helper text looks different in the components that use input internally. For instance, in the Material theme, the helper text in the combo and the input components has margin-top and padding-inline, but there isn't any padding or margin in the select. The helper text color is also different - gray-600 in the select and the combo components and gray-700 in the input.

Fluent and Indigo themes also have some discrepancies regarding the margins and paddings.

## Attachments  
Screen Shot 2023-01-16 at 9 09 22 PM

### LINKED ISSUE #620: Combo / Select toggle icon invisible in disabled state
## Description  
The combo/select suffix container uses one and the same color for the background and the icon color in disabled state so the toggle icon becomes invisible. The bug refers to the Indigo theme only - light and dark variants.

## Attachments  
Screen Shot 2023-01-17 at 11 11 33 AM

simeonoff:
Closes #619
Closes #620


SisIvanova (dismissed):

SisIvanova (approved):

IGSvt:
-The used icon (chevron) should be applied only to the Material theme. The rest of the themes use a caret icon. 
- The suffix in the Material theme has a darker background in case it is enabled, only in the disabled state it has the same background as the input. 
-The Drop-down should not shrink in width once an item has been picked from the list 
-The suffix container should be 48px in width and take all the available space in height

SisIvanova (approved):

PR #676, opened 2023-02-22
chore: Fixed Intl unit test & updated dependencies

rkaraivanov:
See https://github.com/nodejs/node/issues/45938


simeonoff:
@rkaraivanov Perhaps it would be prudent to wait for the Vite PR that introduces changes to the package.json file before we update the dependencies. It would be fine if this PR introduces changes to the tests only.

simeonoff (approved):

PR #711, opened 2023-03-22
fix(tabs): make tabs and tab panels to take the full width of their parent

### LINKED ISSUE #710: Components placed inside tab are not stretching vertically
## Description  
When components like Grid or Chart are placed inside a Tab Layout they won't stretch vertically. The easiest way to reproduce this is to generate an application with the App Builder OR use the attached sample [untitled-app (21)_webComponents.zip](https://github.com/IgniteUI/igniteui-webcomponents/files/11037646/untitled-app.21._webComponents.zip)

## Steps to reproduce  

1. Open the attached app or create a [simple app with the AB](https://appbuilder.indigo.design/app/6gie6fg3vmdo/preview)
2. Run it
3. Observe

## Result  
image

## Note
This style configuration is working in Angular Tab Component

## Expected result  
Components set with the style configuration below to take the full vertical available space.

```
:host {
      height: 100%;
      display: flex;
      justify-content: flex-start;
      align-items: stretch;
      align-content: flex-start;
    }
    .tabs {
      min-width: 0;
      min-height: 0;
      flex-grow: 1;
      flex-basis: 0;
    }
    .row-layout {
      display: flex;
    }
    .group {
      justify-content: flex-start;
      align-items: stretch;
      align-content: flex-start;
      overflow: auto;
      position: relative;
      min-width: 100%;
      min-height: 50px;
    }
    .tab-item-content {
      height: 100%;
      min-width: 100%;
      min-height: 50px;
    }
    .grid {
      min-width: 600px;
      min-height: 300px;
      flex-grow: 1;
      flex-basis: 0;
    }
    .category-chart {
      min-width: 400px;
      min-height: 300px;
      flex-grow: 1;
      flex-basis: 0;
    }
```

desig9stein:
Closes #710

### DIFF
--- a/src/components/tabs/themes/light/tab-panel.base.scss
+++ b/src/components/tabs/themes/light/tab-panel.base.scss
@@ -2,8 +2,9 @@
 @use '../../../../styles/utilities' as *;
 
 :host {
-    flex-grow: 1;
-    overflow: hidden;
+    overflow-x: hidden;
+    overflow-y: auto;
+    flex: 1 1 auto;
 }
 
 :host(:not([hidden])) {
--- a/src/components/tabs/themes/light/tab-panel.base.scss
+++ b/src/components/tabs/themes/light/tab-panel.base.scss
@@ -2,6 +2,10 @@
 @use '../../../../styles/utilities' as *;
 
 :host {
+    height: 100%;
     overflow: hidden;
-    flex-grow: 1;
+}
+
+:host(:not([hidden])) {
+    display: block;
 }
--- a/src/components/tabs/themes/light/tab-panel.base.scss
+++ b/src/components/tabs/themes/light/tab-panel.base.scss
@@ -2,7 +2,7 @@
 @use '../../../../styles/utilities' as *;
 
 :host {
-    height: 100%;
+    flex-grow: 1;
     overflow: hidden;
 }
 
--- a/src/components/tabs/themes/light/tabs.base.scss
+++ b/src/components/tabs/themes/light/tabs.base.scss
@@ -6,8 +6,7 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 :host {
     display: flex;
     flex-direction: column;
-    height: 100%;
-    width: 100%;
+    flex: 1;
 }
 
 :host([alignment='start']) [part='headers-scroll'] {
@@ -45,6 +44,7 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
     overflow: hidden;
     min-height: rem(42px);
     background: color(surface, 500);
+    flex: none;
 }
 
 [part='headers-content'] {
@@ -65,6 +65,8 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 }
 
 [part='content'] {
+    display: flex;
+    flex-direction: column;
     flex-grow: 1;
 }
 
--- a/src/components/tabs/themes/light/tabs.base.scss
+++ b/src/components/tabs/themes/light/tabs.base.scss
@@ -4,7 +4,10 @@
 $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 
 :host {
-    display: block;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    width: 100%;
 }
 
 :host([alignment='start']) [part='headers-scroll'] {
@@ -59,6 +62,10 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 [part='headers-scroll'] {
     display: flex;
     height: 100%;
+}
+
+[part='content'] {
+    flex-grow: 1;
 }
 
 [part='selected-indicator'] {
--- a/src/components/tabs/themes/light/tabs.base.scss
+++ b/src/components/tabs/themes/light/tabs.base.scss
@@ -6,7 +6,6 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 :host {
     display: flex;
     flex-direction: column;
-    flex: 1;
 }
 
 :host([alignment='start']) [part='headers-scroll'] {
--- a/src/components/tabs/themes/light/tabs.base.scss
+++ b/src/components/tabs/themes/light/tabs.base.scss
@@ -6,6 +6,7 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 :host {
     display: flex;
     flex-direction: column;
+    overflow: hidden;
 }
 
 :host([alignment='start']) [part='headers-scroll'] {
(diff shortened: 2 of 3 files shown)


simeonoff (approved):

viktorkombov:
This PR will not resolve the issue #710, because when an app is generated with App Builder, the content of the tabs will be put in an additional container, which height is not set to 100%. So, I believe that the issue should be resolved in the App Builder repo.

Untitled

zdrawku:
@viktorkombov is this specifically for the WC/Blazor code generation? For Angular this is not happening

zdrawku:
We will add  100% height to the wrapping container (AB change) and upon successful testing I will merge the PR

pmoleri:
Hi @desig9stein, I'd be good to compare your changes to [igniteui-angular !11515]https://github.com/IgniteUI/igniteui-angular/pull/11515/files](https://github.com/IgniteUI/igniteui-angular/pull/11515/files).

Those changes allowed the Angular Tabs component to be used either stretched or accommodate to the content. Perhaps there's something useful to consider here too.

src/components/tabs/themes/light/tabs.base.scss
```diff
@@ -4,7 +4,10 @@
 $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 
 :host {
-    display: block;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    width: 100%;
```
pmoleri, line 10: Using height/width: 100% looks like a problem waiting to happen.
This should be decision of the owner of Tabs component.

src/components/tabs/themes/light/tabs.base.scss
```diff
@@ -61,6 +64,10 @@ $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
     height: 100%;
 }
 
+[part='content'] {
```
pmoleri, line 67: I think the counter part of this is setting `part='headers'` to `flex: none` = 0 0 auto;

pmoleri:
> We will add 100% height to the wrapping container (AB change) and upon successful testing I will merge the PR

@zdrawku I don't understand this.
I don't see the extra container here:
image

viktorkombov:
> > We will add 100% height to the wrapping container (AB change) and upon successful testing I will merge the PR
> 
> @zdrawku I don't understand this. I don't see the extra container here: image

@pmoleri , currently we are creating this extra container only for WC. So, after we disscussed this with @zdrawku , we decided to do this and for Blazor in order to not styling igc-tab-panel directly:

image

src/components/tabs/themes/light/tab-panel.base.scss
```diff
@@ -2,6 +2,10 @@
 @use '../../../../styles/utilities' as *;
 
 :host {
+    height: 100%;
```
pmoleri, line 5: Maybe:
`part='content'` should be flex
and tab-panel should be `1 1 auto` like in [igniteui-angular !11515](https://github.com/IgniteUI/igniteui-angular/pull/11515/files).

src/components/tabs/themes/light/tabs.base.scss
desig9stein: Since the issue those not specify if the fix should be only for tabs that are inside a `flex` container, using `height:100%` and `width:100%` ensures that the fix will work even if the tabs are not inside a `flex` container. If that is not a requirement I can do it with `flex-grow: 1`;

src/components/tabs/themes/light/tab-panel.base.scss
desig9stein: making tab panel `flex: 1 1 auto` will require the content to be `display: flex`, not sure if that is OK.

src/components/tabs/themes/light/tabs.base.scss
desig9stein: What are your concerns, do you see a use case where this can have harmful effect?

pmoleri: - Width: you already declared tabs to be "flex", that's block by default so width: 100% is implicit.
- 2 examples of harmful effects:
  1. when in a flex layout it changes the auto basis, so if you want to put a tabs and a div side by side, the tabs will use all the space, e.g.
     
    image

  2. when in a regular layout, if you put a tabs and a button side by side it will make the whole layout overflow
     e.g.
     image

Why having different styling for WC and Angular? After all the refinement Angular has gone through, we start from scratch only to repeat the same mistakes or different ones?

src/components/tabs/themes/light/tab-panel.base.scss
pmoleri: Angular component was implemented with flex end to end, because it has the benefit of being able to work in both scenarios:
 - accommodate to content
 - stretch to parent layout

Depending on what you want. When not using flex, usually you can get only one of them working correctly.

You can test the bug of https://github.com/IgniteUI/igniteui-angular/issues/11479 with this implementation and see if it handles it properly.

src/components/tabs/themes/light/tabs.base.scss
pmoleri: What if I want bigger headers? i.e. put header content that uses more than 42px?

IMO this is the most used flex rule, you want one fixed part and one flexible part, you use `flex: none` and `flex: 1` works like a charm 99% of the time.

pmoleri: Correction, flex: none / flex: 1 -> when you only want to fill the available space.
In this case we want content size to be considered to, so it's flex: none / flex: 1 1 auto, which I believe it's equivalent to the flex-grow: 1 that you have here.

simeonoff (dismissed):

simeonoff:
@zdrawku Can you validate this PR?

src/components/tabs/themes/light/tabs.base.scss
```diff
@@ -4,7 +4,9 @@
 $tabs-animation-function: cubic-bezier(.35, 0, .25, 1);
 
 :host {
-    display: block;
+    display: flex;
+    flex-direction: column;
+    flex: 1;
```
pmoleri, line 9: this equals to `1 1 0`, meaning that the `auto` is lost, I think setting only `flex-grow: 1` is a better alternative.

But, in any case, I don't understand why using this rule here which is not in the Angular component, that would mean different default behavior across platforms.

@simeonoff FYI, because we talked about this too.

desig9stein: I am adding this to make the tabs component stretch by default, however, I just saw that the app builder is adding the `1 1 0` by default, so I will remove it.

pmoleri (dismissed):
From code perspective the changes looks good.
I haven't done any testing though.

simeonoff (dismissed):

simeonoff:
@zdrawku We don't anticipate any more code changes. You can safely test and verify if the solution works for your use case.

zdrawku:
@simeonoff thanks, @viktorkombov is doing final testing today.

PR #716, opened 2023-03-28
fix(dialog): use keepOpenOnEscape instead of closeOnEscape

### LINKED ISSUE #715: Dialog - closeOnEscape has default value true
Dialog - `closeOnEscape` has default value `true`. Replace with new property that has default value `false`.

igdmdimitrov:
Closes #715


damyanpetev (approved):

igdmdimitrov:
@gmurray81, we had to deprecate one property of the Dialog in favor of a new one.  Is there anything else I should do in order for this change to propagate to Blazor?

PR #724, opened 2023-04-04
fix(select): prefixes/suffixes and helper text not being rendered

### LINKED ISSUE #722: Select - Slots need additional render cycle to appear
## Description  
Prefix, Suffix & other slots in the Select are not rendered initially. They appear after you click on the element.

## Steps to reproduce  

1. Open [Codesandbox sample](https://codesandbox.io/s/mystifying-panini-ns04sh?file=/index.html)

## Result  
Prefix, Suffix & other slots' content is not visible. It becomes visible after clicking on the element.

## Expected result  
Content of the slots should be visible

simeonoff:
Closes #722


src/components/select/select.ts
```diff
@@ -252,6 +252,14 @@ export default class IgcSelectComponent extends EventEmitterMixin<
     super.firstUpdated();
     await this.updateComplete;
 
+    if (this.hasPrefixes || this.hasSuffixes) {
+      this.requestUpdate();
+    }
+
+    if (this.helperText) {
+      this.requestUpdate();
+    }
```
igdmdimitrov, line 261:
```suggestion
    if (this.hasPrefixes || this.hasSuffixes || this.helperText) {
      this.requestUpdate();
    }
```

rkaraivanov: Maybe attach a general `slotchange` handler to the slots which re-renders the component when there is slotted content?

rkaraivanov (approved):

PR #728, opened 2023-04-06
Implement Element Internals

### LINKED ISSUE #726: Implement ElmentInternals for appropriate elements
Since all relevant browsers support the relevant parts of the [ElementInternals](https://html.spec.whatwg.org/multipage/custom-elements.html#the-elementinternals-interface) specification,
it is time to implement it for our relevant form components.

image

rkaraivanov:
Closes #726


kdinev:
@rkaraivanov Could you please resolve conflicts?

rkaraivanov:
@desig9stein 
Can you go over the merged changes? I think the switch component invalid style is borked currently

mddragnev:
Single selection combo does not clear its state(the selected value) and the form values  corresponding to the component when the form is being reset. You can see the console that if a submit button is clicked then the form data contains values for both controls.
On the following image is a default and simple combos where both of them have some selected values.
Screenshot 2023-08-02 at 13 12 26
Once you click reset the combos looks like this. You can see that the simple combo still contains value. If you click once more the submit button then the formdata still contains the simple combo value (i think it should not after reset is clicked). You can see this in the console of the second image
Screenshot 2023-08-02 at 13 13 58

rkaraivanov:
@mddragnev 
Should be fixed now

mddragnev:
If u have a dropdown with an input like in the following image:
Screenshot 2023-08-03 at 14 25 20
This is inside a form: (modified the sample from the dropdown form story)
``` 
<form action="" @submit=${formSubmitHandler}>
  <div>
  <igc-dropdown
    id="ddlCountry"
    @igcChange=${(_ev: CustomEvent) => {
      (document.getElementById('txtCountry') as IgcInputComponent).value = (
        _ev.detail as IgcDropdownItemComponent
      ).value;
    }}
    .scrollStrategy=${scrollStrategy}
  >
    <igc-input
      slot="target"
      type="text"
      label="Country"
      id="txtCountry"
      size="small"
      style="width: 150px"
    ></igc-input>
    <igc-dropdown-group>
      <igc-dropdown-header slot="label">Europe</igc-dropdown-header>
      ${countries
        .slice(0, 2)
        .map(
          (item) => html`<igc-dropdown-item>${item}</igc-dropdown-item>`
        )}
    </igc-dropdown-group>
    <igc-dropdown-group>
      <igc-dropdown-header slot="label">North America</igc-dropdown-header>
      ${countries
        .slice(2, 4)
        .map(
          (item) => html`<igc-dropdown-item>${item}</igc-dropdown-item>`
        )}
    </igc-dropdown-group>
    <igc-dropdown-group>
      <igc-dropdown-header slot="label">Asia</igc-dropdown-header>
      ${countries
        .slice(4)
        .map(
          (item) => html`<igc-dropdown-item>${item}</igc-dropdown-item>`
        )}
    </igc-dropdown-group>
  </igc-dropdown>
</div>
  ${formControls()}
</form>
```
Then u are able to submit the form but the form data seems to be empty. Anyway, the reset button works as expected and clears the selected input value. (Did not test it if it clears the form data correct because it is not populated yet)

rkaraivanov:
@mddragnev
Can you set a **name** attribute on the igc-input in your dropdown sample and try again?
Form controls implicitly [require](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name) the attribute otherwise the input control is not submitted.

mddragnev:
@rkaraivanov Yup that was my bad. Didnt see that it is missing. Works correct.

mddragnev:
If you look at the maskinput form story and observe the behavior of the mask pattern validation input (the last one)
```
 <igc-mask-input
          name="required-mask-pattern"
          label="Mask pattern validation"
          mask="(+35\\9) 000-000-000"
        ></igc-mask-input>
```
If you make the input invalid because of bad Input (the mask is not full) as per mask-input definitions the message should be `All required positions should be filled`. Although the message is `This field is required`
At the following image you can see that the validity object is correct and it says that the input is bad but the text message is wrong.
Screenshot 2023-08-03 at 15 14 53
I think that the problem is inside the `updateValidity` methon inside the form mixin.

mddragnev:
Also, for if you have the following configuration of the mask input
```
        <igc-mask-input
          required
          label="Formatted value mode"
          name="mask-formatted-value"
          mask="(CCC) (CCC)"
          value-mode="withFormatting"
          value="123456"
        ></igc-mask-input>
```
It looks like this
Screenshot 2023-08-03 at 15 37 02
Note that it has a default value.
If you reset it (from the reset button of the form) then the mask parser is behaving weird and the result is 
Screenshot 2023-08-03 at 15 38 24

Even though, the reset button is working correcly (reseting the form and setting the default value for the input) will you look at this or should I log it as a separate issue?

mddragnev:
At the radio group storybook in the last fieldset the radio group (all of the radios) are required. If you try to submit it doesnt let you resulting in the following behavior (all radios becomes invalid - in red). 
Screenshot 2023-08-03 at 16 36 38
If you select a radio and then reset the form and try to submit it then only the previous selected radio becomes red:
Screenshot 2023-08-03 at 16 37 23

rkaraivanov:
> Something that is a bit weird to me about the rating inside a form is that the components default value is 0 which results in something like this: Screenshot 2023-08-03 at 16 49 53 There is only 1 rating with selected value but the form data contains all of them. I understand that this is a default value and is probably by specification but just wanted to point it out because other components have the ability not to have a default value (then submitting the form would not cotain them as data)

@mddragnev 
Yeah, the rating component mimics (with some differences) the native
`<input type="range" />` element. As such, it always has a default *valid* value even when the user has not set a value or has passed
an invalid one. So as long as the rating has a **name** attribute set
and it is not **disabled**, it will be submitted.

@damyanpetev and I had a chat on whether the **required** attribute
should be applicable to the element. Both of us are inclined to leave it as it is for now. That is the user has to explicitly set the attribute and any value different than 0 will pass the validation.

Which means it will need additional styling as it currently has none.

@simeonoff @sdimchevski

mddragnev:
If you look at the select story you can see that the first two selects have default values ( the first through value input of the select and the second through the selected input of the select item).
Screenshot 2023-08-04 at 12 57 30

Lets say that we change both of those default values to something else. Then we decide to reset the from which contains both selects. In my understanding both select components should reset to their default value (from the image above both should have value Bulgaria) which does not happen for the second select which has its default value through the selected item. On the other hand, it does make some sense that reseting the form actually clears all selected items but leaves default value only on those select components that implicitly has `value` input set. @rkaraivanov Let me know your thoughts on this

mddragnev:
Trying to submit a required switch which is not on does not have any error state indication.

rkaraivanov:
@mddragnev 
For the select it is an issue, since the default browser select element restores based on the initial selected attribute. I pushed a fix for that.

As for the switch, I mentioned that to @desig9stein several weeks ago, and what I got back is that it is by design.

@sdimchevski @desig9stein 
Can you confirm that this behavior is expected?

mddragnev (approved):

desig9stein:
> Trying to submit a required switch which is not on does not have any error state indication.

The switch does not provide any visual indication of the error state by default, however, since we provide variables for the error state the users are able to change them if they want to.

simeonoff (approved):

src/components/combo/combo.ts
```diff
@@ -441,9 +432,7 @@ export default class IgcComboComponent<T extends object = any>
       this.selectionController.select(this._value as Item<T>[]);
     }
 
-    if (!this.singleSelect) {
-      this.updateValue();
-    }
```
simeonoff, line 446: Removing this check regresses the single selection search behavior. To see the regression in action in single selection mode type a search term, say "Sofia", press enter to select the matched item. Now while still focused, type "v" to search for "Varna". The letter "v" is not displayed in the input field.

rkaraivanov: Can you re-test it with the latest changes?

ChronosSF (approved):

PR #729, opened 2023-04-07
Fix nested tabs selection

### LINKED ISSUE #713: An igc-tabs inside an igc-tab-panel (nested tab components) doesn't work well
## Description  

An `<igc-tabs>` inside an `<igc-tab-panel>` (nested tab components) doesn't work well.

## Steps to reproduce  

1. Put `<igc-tabs>` component inside another `<igc-tabs>` and its `<igc-tab-panel>` component, like the following code.

```html
<igc-tabs>
    <igc-tab panel="home">
        Home
    </igc-tab>
    ...
    <igc-tab-panel id="home">

        <igc-tabs>
          <igc-tab panel="P1">P1</igc-tab>
          ...
          <igc-tab-panel id="P1">
              Panel 1
          </igc-tab-panel>
          ...
        </igc-tabs>

    </igc-tab-panel>
    ...
```

3. Run it and click the tab in  the tab panel.

## Result  

The child tab contents disappeared.

movie-002

## Expected result  

An active tab and its contents should be switched correctly.

## Attachments  

You can reproduce this issue with the project attached below. (After downloading and extracting it, executing `npm i` and `npm start` will launch the app.)

[📦nested-igc-tabs.zip](https://github.com/IgniteUI/igniteui-webcomponents/files/11058857/nested-igc-tabs.zip)

mtsvyatkova:
Closes #713

### DIFF
--- a/src/components/tabs/tabs.ts
+++ b/src/components/tabs/tabs.ts
@@ -198,7 +198,8 @@ export default class IgcTabsComponent extends EventEmitterMixin<
       const attributes = getAttributesForTags<IgcTabComponent>(
         records,
         'igc-tab'
-      );
+      ).filter((e) => e.closest(this.tagName)?.isSameNode(this));
+
       const changed = getNodesForTags<IgcTabComponent>(
         records,
         this,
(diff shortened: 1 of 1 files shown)


rkaraivanov (commented):
This will work, however it may be better to handle that check at the observer
callback level.

```typescript
const attributes = getAttributesForTags<IgcTabComponent>(
        records,
        'igc-tab'
      ).filter((e) => e.closest(this.tagName)?.isSameNode(this));
```

Doing this will keep the changed tag collection only to the relevant parent
tabs component (other ancestors will receive empty collections).

rkaraivanov (approved):

PR #756, opened 2023-04-28
feat(tree): animate component

### LINKED ISSUE #846: Animations: Tree component
Animate the tree component in a fashion similar to the one in Ignite UI for Angular.

SisIvanova:
Closes #846
Related #610


simeonoff (approved):

simeonoff:
@SisIvanova There's a slight jump when expanding a leaf nodes if sizes other than large are used. To reproduce, change the size to medium or small, expand Tree Node 1 and then expand Tree Node 1.1.

src/components/tree/tree-item.ts
```diff
@@ -82,6 +88,10 @@ export default class IgcTreeItemComponent extends LitElement {
   @blazorSuppress()
   public wrapper!: HTMLElement;
 
+  /** @private */
+  @query('[role="group"]', true)
+  public group!: HTMLElement;
```
rkaraivanov, line 93: Do this need to be a public property at all?

rkaraivanov (changes requested):

simeonoff (dismissed):

simeonoff (approved):

src/components/tree/tree-item.ts
simeonoff: This has been addressed.

rkaraivanov (approved):

PR #760, opened 2023-05-04
fix(stepper): differences with igniteui-angular

desig9stein:


simeonoff:
@desig9stein What's the state of this PR?

desig9stein:
The stepper in webC follows the latest design spec, we can update angular to match the webC design.

PR #779, opened 2023-06-01
Refactor Input, select and combo to use colors from schemas

### LINKED ISSUE #767: Style the input components (Input, Select, Combo, etc.) using schemas from theming

desig9stein:
closes #767

- Move all colors inside the shared base file to individual themes.
- Prepare all themes to use schema colors
- Make input material theme consume colors from the schema

### DIFF
--- a/src/components/input/themes/dark/input.dark.bootstrap.scss
+++ b/src/components/input/themes/dark/input.dark.bootstrap.scss
@@ -1,7 +1,17 @@
+@use 'sass:map';
+@use '../../../../styles/utilities' as *;
+@use '../../../../styles/themes/schemas' as *;
+
+$schema: extend(
+    (
+        name: 'ig-input-group'
+    ),
+    map.get($dark-bootstrap-schema, 'input-group')
+);
+$theme: digest-schema($schema);
+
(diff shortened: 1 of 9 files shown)


src/components/select/select.ts
```diff
@@ -488,7 +489,14 @@ export default class IgcSelectComponent extends EventEmitterMixin<
...
             <slot name="suffix" @slotchange=${this.inputSlotChanged}></slot>
           </span>
-          <span slot="suffix" part="toggle-icon" style="display: flex">
+          <span
+            slot="suffix"
+            part="${partNameMap({
+              'toggle-icon': true,
+              filled: this.value!,
+            })}"
+            style="display: flex"
```
simeonoff, line 498: Do we need to to this inline?

simeonoff (changes requested):
- Wrong toggle button background in combo dark theme
- Wrong input text color in light theme in indigo theme
- Wrong toggle button background color in indigo theme in light and dark themes
- The floated label jumps up and down in the multi-select combo material box variant on focus when no value is selected

simeonoff (approved):

PR #785, opened 2023-06-07
fix(combo): parent overflow issues

### LINKED ISSUE #781: IgcGrid + IgcCombo integration scenario issues
## Description  
When combo is templated inside the igcGrid cell's there are some issues that arise from the combo list items rendering inside the grid, main ones being:
-  Combo's list items get cut-off (cannot go outside the grid's display container since it does not show overflows):
image
- Grid's Resize observer gets triggered a lot both on initial rendering and on any combo interaction:
image

Need some way to render and position list items outside of the grid (maybe in overlay like in angular's combo).

## Steps to reproduce  

1. Run sample: 
[combos.zip](https://github.com/IgniteUI/igniteui-webcomponents/files/11662516/combos.zip)

## Result  
Issues with combo content getting cut-off and resize observer triggering on any combo render and interaction.

## Expected result  
Combo rendering with no cut-off and not triggering resize observer so much. Preferably working like in the [angular combo](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/cascading-combos#angular-grid-with-cascading-combos-sample-overview) when integrated inside the grid. 

## Attachments  

simeonoff:
Closes #781

When the combo is placed in a parent with `overflow: hidden` set to it, the drop-down list of options gets cut off.


MayaKirova:
Unfortunately it does not fix it for the grid scenario. The grid virtualization containers as well as rows/cells also have `position: relative` on them so the scenario is more like:
```
<div style="overflow:hidden; position: relative; top:0px; height:80px;">
<igc-combo position-strategy="fixed" placeholder="Choose Country..." single-select></igc-combo>
</div>

```
In which case it still gets cut-off:
image

src/components/combo/combo.ts
```diff
@@ -325,6 +325,18 @@ export default class IgcComboComponent<T extends object>
     return html`${this.groupKey && item[this.groupKey]}`;
   };
 
+  /**
+   * Sets the component's positioning strategy.
+   */
+  @property({ attribute: 'position-strategy' })
+  public positionStrategy: 'absolute' | 'fixed' = 'fixed';
```
damyanpetev, line 332: This doesn't seem to work or do much at least. This bit I assume is important:
> Ensure your initial layout matches the strategy:
#tooltip { position: fixed; ....
https://floating-ui.com/docs/computePosition#strategy

So here:
https://github.com/IgniteUI/igniteui-webcomponents/blob/94c199bd5c91a048cd1ad098f56a0b2bddf6a489/src/components/combo/combo.ts#L867-L872
You need something like the dropdown does:
https://github.com/IgniteUI/igniteui-webcomponents/blob/94c199bd5c91a048cd1ad098f56a0b2bddf6a489/src/components/dropdown/dropdown.ts#L523-L525
image

src/components/combo/combo.ts
```diff
@@ -325,6 +325,18 @@ export default class IgcComboComponent<T extends object>
...
+
+  /**
+   * Whether the dropdown's width should be the same as the target's one.
+   */
+  @property({ type: Boolean, attribute: 'same-width' })
+  public sameWidth = true;
```
damyanpetev, line 338: This can actually be a bit of an issue - boolean flag true by default I believe is something we avoid, no?

damyanpetev (changes requested):

src/components/combo/combo.ts
simeonoff: Ah, yes. 👍

src/components/combo/combo.ts
```diff
@@ -325,6 +326,19 @@ export default class IgcComboComponent<T extends object>
     return html`${this.groupKey && item[this.groupKey]}`;
   };
 
+  /**
+   * Sets the component's positioning strategy.
+   */
+  @property({ attribute: 'position-strategy' })
+  public positionStrategy: 'absolute' | 'fixed' = 'fixed';
```
rkaraivanov, line 333:
```suggestion
  public positionStrategy: 'absolute' | 'fixed' = 'absolute';
```
Can we keep the old behavior?

src/components/combo/combo.ts
```diff
@@ -325,6 +326,19 @@ export default class IgcComboComponent<T extends object>
...
+
+  /**
+   * Whether the dropdown's width should be the same as the target's one.
+   * True by default.
+   */
+  @property({ type: Boolean, attribute: false })
+  public sameWidth = true;
```
rkaraivanov, line 340:
```suggestion
  /**
   * Whether the dropdown's width should be the same as the target's one.
   * True by default.
   */
  @property({ type: Boolean, attribute: 'same-width' })
  public sameWidth = false;
```
This should match https://github.com/IgniteUI/igniteui-webcomponents/blob/0a24a42356476bff4dc16df8863792a523905bc9/src/components/select/select.ts#L188-L193

rkaraivanov (changes requested):
Can you also bind the strategy and same-width properties in the default story,
as currently toggling the controls in the storybook does nothing?

src/components/combo/combo.ts
simeonoff: As far as I understood from @damyanpetev it doesn't work with absolute in the context of the grid!?

src/components/combo/combo.ts
damyanpetev: 👀 Oh we have `sameWidth` true by default on the select? Well that's awkward.
I was going to comment the position strategy should match as well.

rkaraivanov (approved):

damyanpetev (approved):

damyanpetev:
@MayaKirova Not sure if you want to check anything else as well, but with this change the combos are working (better at least) in the grid:
image

**Note**: There are still _limitations_ as even the fixed strategy can't escape the stacking contexts of the grid and the body content has a context reset (`z-index: 1`) so combo/select dropdowns will still clip under grid headers and scrollbars. Really not sure there's anything that can be done for that. Perhaps at least document them somewhere.
Also, with #788 the fixed position behaves as the absolute did before (i.e. tracks its target when scrolling), which is great. There's still an issue with the Grid integration again tho - the scroll container doesn't really emit a scroll event that the floating UI can react to and still might need customizations.

MayaKirova:
@damyanpetev I think it looks good in the context if this sample. 
I can add a note somewhere in the cascading combo's topic regarding the `z-index`: https://github.com/IgniteUI/igniteui-xplat-docs/pull/845
I guess that should be enough.

PR #788, opened 2023-06-09
feat: Floating elements auto-update behavior

rkaraivanov:
`igcToggle` directive by default will try to keep the floating element anchored to its reference element on scrolling and resizing the screen.

### DIFF
--- a/src/components/toggle/toggle.directive.ts
+++ b/src/components/toggle/toggle.directive.ts
@@ -5,6 +5,8 @@ import {
   shift,
   size,
   Middleware,
+  autoUpdate,
+  ComputePositionConfig,
 } from '@floating-ui/dom';
 import {
   directive,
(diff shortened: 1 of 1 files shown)


src/components/toggle/toggle.directive.ts
```diff
@@ -20,57 +22,86 @@ import type { IgcToggleController } from './toggle.controller.js';
...
-      placement: options.placement ?? 'bottom-start',
-      strategy: options.positionStrategy ?? 'absolute',
-      middleware: this.createMiddleware(options),
-    }).then(({ x, y }) => {
-      Object.assign(this.floatingElement.style, {
-        left: 0,
-        top: 0,
-        transform: `translate(${x}px,${y}px)`,
-      });
+    Object.assign(this._floatingElement.style, {
```
damyanpetev, line 38: Kind of an odd place to have this, assuming from the old code only `updatePosition` needs to switch open state _once_. Probably no noticeable hit to speak of, but still odd. Also bit funky the element needs to be cached, considering the part appears to be a static constructor param so likely just property access `this.part.element` (with type asserts) can do just fine

damyanpetev (approved):
Very minor nit for the `floatingElement` getter, otherwise LGTM and can go in as is.

PR #801, opened 2023-06-21
feat(components): consume border radius from component schemas

### LINKED ISSUE #805: Themes: Use border radius in components from schemas
The radius factor should be consumed from the schema files for each component.

SisIvanova:
Closes #805 
Related to IgniteUI/igniteui-angular/issues/13108


src/components/button/themes/button/light/button.bootstrap.scss
```diff
@@ -16,15 +16,15 @@ $outlined-theme: digest-schema($outlined-schema);
     @include css-vars($flat-theme);
 
     [part='base'] {
-        @include border-radius(var-get($flat-theme, 'border-radius'));
+        @include border-radius(map.get($flat-theme, 'border-radius')...);
```
simeonoff, line 19: This way of implementing it completely removes the ability for the user to set a custom border-radius via the CSS variable.

I believe the following should still work.
```suggestion
        @include border-radius(var-get($flat-theme, 'border-radius')...);
```

simeonoff (commented):
Don't replace var-get with map.get as this would remove the ability to use custom CSS properties for modifying the border-radius.

simeonoff (changes requested):
- Button
    - All buttons have a squared-off radius in their dark themes
- Calendar
    - Vertical calendars need more attention when it comes to rounding off various corners
- Checkbox
    - Border always square in all dark themes
- Combo & Select
    - Dark themes always square
    - Toggle button border bottom right radius is rounded with `--ig-radius-factor` set to 1 in material box variant
    - Toggle button border top right radius is square with `--ig-radius-factor` set to 1 in indigo
- Input 
     - Dark themes always square
- Stepper 
    -  Compare against Angular to ensure visual parity.
- Switch 
     - Always square in the dark themes.

simeonoff (approved):

PR #808, opened 2023-06-27
chore: igc-select flip on by default

rkaraivanov:
Hide deprecated properties from elements manifest and API documentation


damyanpetev (commented):
Might want to note the deprecation in the changelog. unless that's a secondary step when releasing? Otherwise, LGTM

damyanpetev (approved):

PR #811, opened 2023-06-27
fix(combo): Do not coerce value to string

### LINKED ISSUE #810: Combo converts value keys to string
## Description  
I am using a combo as an editor in the grid. Here is how I do this:
```js
  public comboInlineEditorTemplate = (ctx: IgcCellTemplateContext) => html`
    <igc-combo .data="${this.friends}" .value="${ctx.cell.editValue}" @igcChange="${(e: CustomEvent<IgcComboChangeEventArgs>) => ctx.cell.editValue = e.detail.newValue}"
        value-key="id"
        display-key="first_name"
        master_view-scope
        class="combo">
    </igc-combo>
  `

My data has this structure:

```ts
export interface Friend {
	id: number;
	first_name: string;
	last_name: string;
}
```

What happens is when combo get the value from the grid cell it is an array of numbers. When `igxChange` fires `e.detail.newValue` returns an array of string. This effectively breaks the selection in the combo as well as the editing in the grid.

rkaraivanov:
Closes #810

### DIFF
--- a/src/components/combo/types.ts
+++ b/src/components/combo/types.ts
@@ -20,6 +20,7 @@ export type ComboChangeType = 'selection' | 'deselection' | 'addition';
 export type ComboRenderFunction<T extends object> = RenderItemFunction<
   ComboRecord<T>
 >;
+export type ComboValue<T> = T | Values<T>;
 
 export interface FilteringOptions<T extends object> {
   filterKey: Keys<T> | undefined;
@@ -35,10 +36,10 @@ export interface GroupingOptions<T extends object> {
 }
 
 /* marshalByValue */
-export interface IgcComboChangeEventArgs {
-  newValue: string[];
+export interface IgcComboChangeEventArgs<T extends object = any> {
+  newValue: ComboValue<T>;
   /* primitiveValue */
-  items: object;
+  items: T[];
   /* blazorAlternateName: changeType */
   type: ComboChangeType;
 }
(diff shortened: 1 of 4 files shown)


simeonoff (approved):

src/components/combo/types.ts
```diff
@@ -35,10 +36,10 @@ export interface GroupingOptions<T extends object> {
 }
 
 /* marshalByValue */
-export interface IgcComboChangeEventArgs {
-  newValue: string[];
+export interface IgcComboChangeEventArgs<T extends object = any> {
+  newValue: ComboValue<T>;
```
damyanpetev, line 40: Figured why the analyzer turned this into object. Um, this should still be array tho, right?
```suggestion
  newValue: ComboValue<T>[];
```
Otherwise if you type the event args proper (Not a bad idea to begin with):
```ts
it('should fire igcChange selection type event on mouse click', async () => {
      const eventSpy = sinon.spy(combo, 'emitEvent');
      const args: Partial<CustomEvent<IgcComboChangeEventArgs<City>>> = {
        cancelable: true,
        detail: {
          newValue: ['BG02'],
          items: [cities[1]],
          type: 'selection',
        },
      };
```
It's really not happy :)
image

damyanpetev: Yup, just noticed, same as `value` already is.

damyanpetev (changes requested):

src/components/combo/types.ts
```diff
@@ -35,10 +36,10 @@ export interface GroupingOptions<T extends object> {
...
   /* primitiveValue */
```
damyanpetev, line 41: @gmurray81 This I'm assuming is likely either wrong or not correct anymore? Items were not primitive values even before, they are actual data items, always. Which would mean they'd likely be tracked and this can be dropped?

gmurray81: I think we might be treating a ref to an item as a primitive value. I'd have to check.

PR #840, opened 2023-07-25
fix(ripple): icon button roundness

### LINKED ISSUE #839: Icon button ripple shape doesn't clip to the round button in Material/Indigo themes
## Description  
Per the title - the ripple extends to a square beyond the round shape of the button

## Steps to reproduce  

1. Go to https://igniteui.github.io/igniteui-webcomponents/?path=/story/iconbutton--basic&globals=theme:material
2. Click on the first the buttons

## Result  
wc-icon-button-ripple

Same is true for the `flat` variant, though only clearly visible with the active highlight or on a colored background.

## Expected result  
Comparing with Ignite UI for Angular https://stackblitz.com/edit/uvfnvr

angular-icon-button-ripple

## Attachments  

SisIvanova:
Closes #839

### DIFF
--- a/src/components/button/themes/icon-button/light/icon-button.bootstrap.scss
+++ b/src/components/button/themes/icon-button/light/icon-button.bootstrap.scss
@@ -15,6 +15,10 @@ $theme: digest-schema(extend(
     @include css-vars-from-theme($theme, 'ig-button');
 
     --size: #{sizable(rem(31px), rem(38px), rem(48px))};
+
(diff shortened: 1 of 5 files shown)


simeonoff (changes requested):
Shouldn't this be handled in all themes? Consider the case in which we set `--ig-radius-factor: 1` for all components. The ripple content will look broken in both bootstrap and fluent.

simeonoff (approved):

PR #842, opened 2023-07-25
fix(dropdown): fix active state styles

### LINKED ISSUE #841: Dropdown: focus styles
## Description  
The focus styles should be added on the `[active]` attr, not on the `:focus` state on the items.

## Steps to reproduce  

1. Switch to bootstrap theme
2. Open the dropdown
3. Move the active/focused item using your keyboard.

## Result  
Since the items are never focused but instead use [active] attr we can't see the correct styles.
 
## Expected result  
The item with attr [active] should have the styles for focus

desig9stein:
Closes #841

### DIFF
--- a/src/components/dropdown/themes/light/item/dropdown-item.fluent.scss
+++ b/src/components/dropdown/themes/light/item/dropdown-item.fluent.scss
@@ -25,7 +25,7 @@ $theme: digest-schema(extend(
     color: var-get($theme, 'hover-item-text-color');
 }
 
-:host(:focus) {
+:host([avtive]) {
     background: var-get($theme, 'focused-item-background');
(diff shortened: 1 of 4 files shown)


src/components/dropdown/themes/light/item/dropdown-item.fluent.scss
```diff
@@ -25,7 +25,7 @@ $theme: digest-schema(extend(
     color: var-get($theme, 'hover-item-text-color');
 }
 
-:host(:focus) {
+:host([avtive]) {
```
simeonoff, line 28:
```suggestion
:host([active]) {
```

simeonoff (changes requested):

simeonoff (approved):

PR #844, opened 2023-07-27
fix(input): wrong padding in indigo theme

SisIvanova:
Related to IgniteUI/igniteui-angular/issues/13255


simeonoff (dismissed):

desig9stein:
Combo and select components have padding after the toggle icon.

Screenshot 2023-07-31 at 17 04 00

simeonoff (approved):

PR #867, opened 2023-08-08
refactor(rating): update the colors for fluent theme according to the UI kit

### LINKED ISSUE #481: Fluent theme, Rating uses wrong colors
## Description  
There are color differences between our fluent theme and [ms fluent](https://developer.microsoft.com/en-us/fluentui#/controls/web/rating#usage)

desig9stein:
closes #481

### DIFF
--- a/src/components/rating/rating-symbol.ts
+++ b/src/components/rating/rating-symbol.ts
@@ -1,5 +1,5 @@
 import { html, LitElement } from 'lit';
-import { styles } from './rating-symbol.base.css';
+import { styles } from './rating-symbol.base.css.js';
 
 /**
  *
--- a/src/components/rating/rating-symbol.ts
+++ b/src/components/rating/rating-symbol.ts
@@ -1,5 +1,5 @@
 import { html, LitElement } from 'lit';
-import { styles } from './rating-symbol.base.css.js';
+import { styles } from './rating-symbol.base.css';
 
 /**
  *
--- a/src/components/rating/rating.fluent.scss
+++ b/src/components/rating/rating.fluent.scss
@@ -4,6 +4,11 @@ $disabled-color: color(gray, 200) !default;
 $disabled-selected-color: color(gray, 500) !default;
 $disabled-label-color: color(gray, 400) !default;
 
+:host {
+    --symbol-full-color: #{color(gray, 900)};
+    --symbol-empty-color: #{color(gray, 700)};
+}
+
 [part='base'],
 :host {
     gap: rem(6px);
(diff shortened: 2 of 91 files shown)


src/components/rating/rating-symbol.ts
```diff
@@ -1,5 +1,5 @@
 import { html, LitElement } from 'lit';
-import { styles } from './rating-symbol.base.css.js';
```
simeonoff, line 2: This should not be removed, keep the .js extension.

src/components/rating/rating.fluent.scss
```diff
@@ -1,9 +1,12 @@
 @use '../../styles/utilities' as *;
 
-$disabled-color: color(gray, 200) !default;
-$disabled-selected-color: color(gray, 500) !default;
 $disabled-label-color: color(gray, 400) !default;
 
+:host {
+    --symbol-full-color: #{color(warn, 900)};
+    --symbol-empty-color: #{color(gray, 700)};
```
simeonoff, line 7: Shouldn't we reflect this in the schema for the rating component as well?

src/components/rating/rating.ts
```diff
@@ -9,10 +9,10 @@ import { Constructor } from '../common/mixins/constructor.js';
 import { EventEmitterMixin } from '../common/mixins/event-emitter.js';
 import { SizableMixin } from '../common/mixins/sizable.js';
 import { clamp, format, isLTR } from '../common/util.js';
-import { styles } from './rating.base.css.js';
-import { styles as bootstrap } from './rating.bootstrap.css.js';
-import { styles as fluent } from './rating.fluent.css.js';
-import { styles as indigo } from './rating.indigo.css.js';
```
simeonoff, line 15: Do not remove the .js extensions from the import statements.

simeonoff (changes requested):

src/components/rating/rating.fluent.scss
desig9stein: Yes, but we can do that when we consume the schema colors for the raring.

simeonoff (approved):

simeonoff (changes requested):
Even though this PR implements the requested changes, it introduces more issues than it resolves. For instance, the selected and unselected symbols in a disabled rating  virtually have the same color. For templated ratings where the same icon is used for both selected and unselected symbols, it will render the rating completely unreadable. Also, the updated colors for selected and unselected rating symbols are too similar in the non-disabled ratings, which decreases the color contrast and consequently accessibility.

@sdimchevski, please reconsider the requested changes.

simeonoff (approved):

sdimchevski:
@simeonoff It does introduce an issue only when the same icon is used for both selected and unselected symbols but using two identical symbols is not that likely. Usually the idea is to have two different symbols like for example one filled icon and one outlined icon. At least that's what is the assumption in Fluent UI where we actually got these colors from.

Screen Shot 2023-08-15 at 14 02 01

PR #871, opened 2023-08-09
refactor: Deprecated validation properties on several components

rkaraivanov:
This is not a breaking change.

Scheduled for removal in v5.0


rkaraivanov:
There is still some work to do, so the stories generation does not emit deprecated properties.

simeonoff (approved):

PR #876, opened 2023-08-11
fix(input): outlined styling issues

### LINKED ISSUE #875: Input/Outlined is not reproduced as expected when it's not filled
## Description  
The Input/Outlined is not reproduced as expected when not filled in the **Disabled** and **Invalid** state.

## Steps to reproduce  

**Invalid state**
1. Select the Input/Docs from the Sidebar
2. Select Outlined to be True
3. Select Invalid to be True
4. Select Required to be True

**Disabled state**
1. Select the Input/Docs from the Sidebar
2. Select Outlined to be True
3. Select Disabled to be True
4. Select Required to be True

## Result

**Invalid state**
The label background is visible, while the label is used as a placeholder because the input is not filled.
image
image

**Disabled state**
image
image

## Expected result

**Invalid state**
When the input is empty, and there is only a placeholder in the Invalid and Disabled state, the border line shouldn't be interrupted.
image
image

**Disabled state**
image
image

## Attachments  

SisIvanova:
Closes #875


simeonoff (approved):

desig9stein:
@SisIvanova When select and combo has no selected value and no placeholder the problem is still there

simeonoff:
@desig9stein Can you verify again?

simeonoff (approved):

PR #881, opened 2023-08-15
fix(input-group): indigo theme bottom line styling

### LINKED ISSUE #879: Themes: Input/Select/Combo Indigo theme glitches
The Input/Select/Combo components do not look right on hover/focus.

https://github.com/IgniteUI/igniteui-webcomponents/assets/4118303/8ceed166-5a04-4703-8b63-5a4423a8cbd8

desig9stein:
closes #879

### DIFF
--- a/src/components/input/themes/light/input.indigo.scss
+++ b/src/components/input/themes/light/input.indigo.scss
@@ -109,15 +109,6 @@ $transition-duration: .25s;
     }
 }
 
-:host(:focus-within) {
-    [part='prefix'],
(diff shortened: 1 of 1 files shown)


simeonoff (commented):
The only remaining issue I see is the transition of the prefix/suffix background appears to lag behind the container itself. Setting the background to a specified value other than inherit seems to fix the issue, however, I am not sure this is the correct approach.

simeonoff (approved):

PR #884, opened 2023-08-16
stepper: consume colors from schemas

### LINKED ISSUE #868: Stepper: consume colors from schema
## Result  
The stepper component does not consume its colors from the stepper schema.

## Expected result  
The stepper component should consume its colors from the stepper schema.

desig9stein:
closes #868


simeonoff (changes requested):
Overall, it looks good, however there are a few outstanding issues:

- The header title is blue in the bootstrap theme as opposed to being black/white as it is in master and also Ignite UI for Angular.
- The header connector (line) is different in color in the Bootstrap dark variant compared to both master and Ignite UI for Angular.
- The indicator text and border colors are different in the Indigo dark theme compared to both master and Ignite UI for Angular.

simeonoff (approved):

PR #886, opened 2023-08-18
feat(stepper): add animations

### LINKED ISSUE #861: Stepper: Implement Animations

simeonoff:
Closes #861


SisIvanova:
If we want to be consistent with Ignite UI for Angular the default animation duration should be 320ms. There is also a difference in the fade animation - in Ignite UI for Angular there is no leave animation (fade out) but here we animate both the current (fade out) and the next step (fade in).

SisIvanova (approved):

PR #900, opened 2023-09-04
fix(calendar): Rotate the month navigation arrows 180deg if the dir attr is set to rtl

### LINKED ISSUE #893: Calendar: Picker arrows aren't flipped in RTL mode
The picker arrows don't point in the right direction when the calendar is used in RTL mode.

Here's a screenshot:
Screenshot 2023-08-29 at 9 48 05

desig9stein:
closes #893


SisIvanova:
When the orientation of the calendar is set to vertical, the picker arrows should be rotated at 90deg instead of 180deg. See the difference between the vertical calendar in ltr and rtl directions. This is how it should be displayed:
Screenshot 2023-09-05 at 2 09 02 PM

SisIvanova (approved):

simeonoff (approved):

PR #921, opened 2023-09-25
refactor(*): update border-radius usage

simeonoff:


SisIvanova (commented):
Same as IgniteUI for Angular - stepper border-radius is not working

SisIvanova (dismissed):

PR #926, opened 2023-09-28
refactor(toast): use colors from schemas

### LINKED ISSUE #916: Toast component doesn't use the variables from the theming package
## Description  
Toast component doesn't use the variables from the theming package

## Steps to reproduce  

1. Open the Toast component from the Storybook
2. Inspect <igc-toast>

## Result  
Upon inspecting you'll notice that the following variables appear as 'not defined': ig-toast-elevation; --ig-toast-background; --ig-toast-text-color; --ig-toast-border-radius

## Expected result  
It is expected that the Toast component uses the variables from the theming package.

## Attachments  
Screenshot 2023-09-20 at 12 18 10

desig9stein:
closes #916


src/components/toast/themes/dark/toast.bootstrap.scss
```diff
@@ -0,0 +1,9 @@
+@use 'sass:map';
+@use 'styles/utilities' as *;
+@use 'themes' as *;
+
+$theme: $bootstrap;
+
+:host {
+    @include css-vars-from-theme($theme, 'ig-toast');
+}
```
simeonoff, line 9: I don't think this will work. The `:host` selector in the dark stylesheets will only work after [this PR](https://github.com/IgniteUI/igniteui-webcomponents/pull/927/) is merged.

simeonoff (changes requested):
I don't think the dark themes will work with the `:host` selector.

simeonoff (approved):

PR #960, opened 2023-10-24
fix(snackbar): remove unnecessary styles

SisIvanova:
Related to IgniteUI/igniteui-angular/issues/13574

### DIFF
--- a/src/components/snackbar/themes/snackbar.base.scss
+++ b/src/components/snackbar/themes/snackbar.base.scss
@@ -43,23 +43,12 @@
 }
 
 :host([action-text]) {
-    igc-button::part(base) {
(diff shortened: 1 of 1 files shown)


simeonoff (commented):
Are the changes in this PR related to any one of the 4 themes? In other words, does the design diverge from the one in the UI kits and by extension Ignite UI for Angular?

SisIvanova:
> Are the changes in this PR related to any one of the 4 themes? In other words, does the design diverge from the one in the UI kits and by extension Ignite UI for Angular?

Yes, the design of the snackbar button was different than the UI kits.

simeonoff (approved):

PR #965, opened 2023-10-26
fix missing extensions

gmurray81:


gmurray81:
These are just typos and we need builds. Merging.

PR #983, opened 2023-11-10
feat: Refactor toggle directive into popover component

rkaraivanov:


stories/dropdown.stories.ts
```diff
@@ -156,269 +160,251 @@ interface IgcDropdownArgs {
...
+      <igc-button id="2nd" onclick="dropdown.show('2nd')">Second</igc-button>
+      <igc-button id="3rd" onclick="dropdown.show('3rd')">Third</igc-button>
+      <igc-button id="4th" onclick="dropdown.show('4th')">Fourth</igc-button>
+    </div>
+    <igc-input
+      id="input"
+      style="max-width: 15rem"
+      label="Focus me"
+      onfocus="dropdown.show('input')"
+    ></igc-input>
```
gedinakova, line 392: There seems to be an issue with this setup of the buttons and input. None of the dropdowns would open on click. The following error is logged: "dropdown.show is not a function
    at IgcButtonComponent.onclick".

gedinakova (changes requested):

src/components/common/controllers/root-click.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(
+    private readonly host: RootClickControllerHost,
+    private config?: RootClickControllerConfig
+  ) {
+    this.host.addController(this);
+  }
+
+  private addEventListeners() {
+    if (!this.host.keepOpenOnOutsideClick) {
+      document.addEventListener('click', this);
```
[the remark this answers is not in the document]
rkaraivanov: Yeah, I pondered the same question when refactoring this. As it stands now, it is compliant with the way Lit does SSR, as `connectedCallback/disconnectedCallback` is not called in SSR environment, according to [this](https://lit.dev/docs/ssr/authoring/#standard-custom-element-and-litelement). Same goes for the reactive controller hooks.

The whole SSR for web components is very much a WiP on browser vendors level so I'm not 100% sure what the best course of action is. For now, this is a separate unit of work, so migrating to a SSR strategy should be easy when the request for such behavior comes.

src/components/dropdown/dropdown-group.ts
```diff
@@ -26,27 +27,24 @@ export default class IgcDropdownGroupComponent extends LitElement {
...
-    this.setAttribute('role', 'group');
-    this.parent = this.getParent();
-  }
-
-  protected getParent() {
-    return this.closest('igc-dropdown')!;
+  constructor() {
+    super();
+    this._internals = this.attachInternals();
+    this._internals.role = 'group';
```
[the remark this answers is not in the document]
rkaraivanov: AxE is just a tool to check if a certain piece of DOM passes their ruleset of what accounts for accessible. The workaround way mainly for the unit tests that invoke `.to.be.accessible()`.

As for the browsers/screen readers(NVDA in my case) they seem to be fine with it.
Here is the Accessibility Tree view from Edge:
image

simeonoff (dismissed):

stories/dropdown.stories.ts
gedinakova: I actually still reproduce this on the 'Docs' page but no on the story page.

rkaraivanov: Should be fixed now.

gedinakova (approved):

PR #984, opened 2023-11-13
fix(button-group): changed igcSelect/igcDeselect to non-cancelable

### LINKED ISSUE #975: Button group select and deselect events are fired before the buttons are selected and deselected with incorrect selected buttons length
## Description  
The `select` and `deselect` events of the `IgcButtonGroup` component are fired before the buttons are selected and deselected. This is relayed to the same behavior in Ignite UI Angular [here](https://github.com/IgniteUI/igniteui-angular/issues/13580).

## Steps to reproduce  
1. Select one button from the button group component.
2. Check the selected buttons/items and observe that the `select` event is fired but the selected buttons/items `length` is **0**.
3. Deselect the button from the button group component.
4. Check the selected buttons/items and observe that the `deselect` event is fired but the selected buttons/items `length` is **1**.

## Result  
When a button is selected from the button group component, the `select` event is fired before the button is selected and the `length` of the selected buttons/items is **0**. When a button is deselected from the button group component, the `deselect` event is fired before the button is deselected and the `length` of the selected buttons/items is **1**.

## Expected result  
When a button is selected from the button group component, the `select` event should fire and the `length` of the selected buttons should be **1**. When a button is deselected from the button group component, the `deselect` event should fire and the `length` of the selected buttons should be 0.

## Attachments  
No attachments.

RivaIvanova:
Closes #975


rkaraivanov (approved):
@RivaIvanova 

LGTM

Can you update the relevant [specification topic](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Button-Group-Specification#events) before merging this?

PR #1015, opened 2024-01-04
feat(rating): Added property to prevent deselection of current value

### LINKED ISSUE #1014: IgcRatingComponent: Prevent unselecting an already selected symbol.
## Is your feature request related to a problem? Please describe.
In AppBuilder, it's desirable when using the rating component, to be able to prevent the ability to un-select an already selected symbol:

In the current implementation of a feedback form, which uses icon buttons, un-selecting a selected button is not possible:

Feedback rating

If this implementation is to use the Angular Rating Component, un-selecting a selected symbol appears to be possible:

Feedback rating local

## Describe the solution you'd like
I haven't been able to find a property or other way to set a given rating symbol to become readonly after having been clicked on.
It would be a useful feature to be able to configure the rating component in such a way.

## Describe alternatives you've considered
I've tried manually looping through all the rating symbols after a click event and adding a readonly/disabled class with style "pointer-events: none;" to the rating symbol which was clicked, but it still doesn't work for me

## Additional context
AppBuilder is currently using IgniteUI for Angular v16.1.11

rkaraivanov:
Closes #1014


simeonoff (dismissed):

simeonoff:
@rkaraivanov I guess the spec will be updated accordingly?

rkaraivanov:
Yes, after the initial code review has passed.

Issue #1017, opened 2024-01-08
Implement sorting of combo groups for group labels and records


mddragnev:
## Description  
There is an issue within the angular product: https://github.com/IgniteUI/igniteui-angular/issues/12876. According to the new specification of the combo (more in the sorting of the groups) there are some changes that need to be addressed. The combos in both libraries (angular and wc) should behave the same

PR #1026, opened 2024-01-15
feat(Combo): Add support for groupingDirection: none .Remove the sorting of the items when combo has grouping.

mddragnev:
Closing #1017

### DIFF
--- a/src/components/combo/operations/group.ts
+++ b/src/components/combo/operations/group.ts
@@ -46,11 +46,13 @@ export default class GroupDataOperation<T extends object> {
       groupBy(data, (item) => item.value[groupKey] ?? 'Other')
     );
 
-    return groups.flatMap(([group, items]) => {
-      items.sort((a, b) =>
-        this.compareObjects(a.value, b.value, displayKey!, direction)
-      );
+    if (direction !== 'none') {
+      groups.sort((a, b) => {
+        return this.orderBy.get(direction)! * a[0].localeCompare(b[0]);
+      });
+    }
 
+    return groups.flatMap(([group, items]) => {
       items.unshift({
         dataIndex: -1,
         header: true,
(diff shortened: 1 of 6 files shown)


src/components/combo/operations/group.ts
```diff
@@ -46,11 +46,13 @@ export default class GroupDataOperation<T extends object> {
...
 
-    return groups.flatMap(([group, items]) => {
-      items.sort((a, b) =>
-        this.compareObjects(a.value, b.value, displayKey!, direction)
-      );
+    if (direction !== 'none') {
+      groups.sort((a, b) => {
+        return this.orderBy.get(direction)! * a[0].localeCompare(b[0]);
+      });
+    }
```
rkaraivanov, line 53:
```suggestion
    if (direction !== 'none') {
      const orderBy = this.orderBy.get(direction)!;
      groups.sort((a, b) => orderBy * comparer.compare(a[0], b[0]));
    }
```
Similar to the other PR you can use an `Intl.Collator` object here for the comparison.

Also, since items are no longer sorted, you can pretty much remove all the code above, which manages resolving item keys and comparisons.

rkaraivanov (changes requested):
https://github.com/IgniteUI/igniteui-webcomponents/blob/7e541ad987967d53cebc87d6552cc04bd9112a55/src/components/combo/combo.ts#L255-L261

Make sure to add the new member of the **groupSorting** union.
```ts
/**
   * Sorts the items in each group by ascending or descending order.
   * @attr group-sorting
   * @type {"asc" | "desc" | "none"}
   */
  @property({ attribute: 'group-sorting', reflect: false })
  public groupSorting: GroupingDirection = 'asc';
```

rkaraivanov (approved):

PR #1065, opened 2024-02-15
feat(button/icon button): update indigo buttons for new ui kit

didimmova:
Related with [this issue](https://github.com/IgniteUI/igniteui-theming/issues/149)
And here is the igniteui-theming [PR](https://github.com/IgniteUI/igniteui-theming/pull/183)


andiesm813:
I think the ONLY issue i see in the buttons in WC is that the SMALL button should have 10px padding on the left and right. The rest seems ok. 
image

didimmova:
> I think the ONLY issue i see in the buttons in WC is that the SMALL button should have 10px padding on the left and right. The rest seems ok. image

I updated the paddings, you can check those and then if everything's the way it should be, put the design:verified label.

src/components/button/themes/button/shared/button.indigo.scss
```diff
@@ -42,6 +48,10 @@ $outlined-theme: $indigo-outlined;
 
     &:focus {
         border-color: var-get($outlined-theme, 'focus-border-color');
+
+        &:hover {
+            border-color: var-get($fab-theme, 'hover-border-color');
```
SisIvanova, line 53: Are you using the border color from `$fab-theme` instead of `$outlined-theme` on purpose?

didimmova: no, thank you for the comment

PR #1068, opened 2024-02-19
feat(btn-group): implement indigo design changes

SisIvanova:
Related to IgniteUI/igniteui-theming/issues/152


didimmova:
- The right border of the button isn't getting the styles on hover, unless it's the last one (in the focus-hovered state it's working for reference)

image

PR #1071, opened 2024-02-20
feat(dialog): update indigo dialog paddings

didimmova:


SisIvanova (changes requested):
- title - wrong font size and line height + unnecessary margin-right 
- the background should be white instead of surface color
- the elevation level in dark mode should be 23 instead of 24 (may be bug in the UI kit)
- wrong content color in dark mode

PR #1093, opened 2024-03-07
chore: update analyzer attributes

damyanpetev:


rkaraivanov (approved):
LGTM

Can we make the analyzer be aware of component members such as connected/disconnected callbacks, register static call, etc. and forgo the process of manually adding the annotations?

PR #1096, opened 2024-03-08
refactor: Migrate to globalThis references

rkaraivanov:


damyanpetev (dismissed):
Assuming the goal is to improve SSR compat and specifically Lit's SSR Shim patches things on `globalThis` already.

damyanpetev (approved):

PR #1117, opened 2024-03-27
fix(button/icon-button): focus styles and width improvements

### LINKED ISSUE #1107: Themes: Update Button and Button Group focus styles
- Button Group: Active state should be without the focus border (This behaviour is applied only in Fluent and Indigo and should be logged as a separate feature/issue for Material & Bootstrap). This is applicable for the Button component as well.
- Button: When we have an Outlined button and a Contained button, if the content is the same, both button widths should be the same, but now the border in the Outlined buttons adds four additional pixels to the button width. (this behaviour is in all themes, not only Indigo)

_Originally posted by @AnjiManova in https://github.com/IgniteUI/igniteui-webcomponents/issues/1090#issuecomment-2006205959_

SisIvanova:
Closes #1107

### DIFF
--- a/src/components/button/themes/button/shared/button.bootstrap.scss
+++ b/src/components/button/themes/button/shared/button.bootstrap.scss
@@ -14,8 +14,6 @@ $outlined-theme: $bootstrap-outlined;
     --component-size: var(--ig-size, #{var-get($outlined-theme, 'default-size')});
 
     [part='base'] {
-        border: rem(1px) solid var-get($outlined-theme, 'border-color');
-
         &:focus {
             border: rem(1px) solid var-get($outlined-theme, 'focus-border-color');
         }
@@ -32,32 +30,28 @@ $outlined-theme: $bootstrap-outlined;
 
 :host(:not([disabled])[variant='flat']) [part='base'],
 :host(:not(:disabled)[variant='flat']) [part='base'] {
-    &:focus,
-    &:active {
+    &:focus-visible {
         box-shadow: 0 0 0 rem(4px) var-get($flat-theme, 'shadow-color');
     }
 }
 
 :host(:not([disabled])[variant='outlined']) [part='base'],
 :host(:not(:disabled)[variant='outlined']) [part='base'] {
-    &:focus,
-    &:active {
+    &:focus-visible {
         box-shadow: 0 0 0 rem(4px) var-get($outlined-theme, 'shadow-color');
     }
 }
 
 :host(:not([disabled])[variant='contained']) [part='base'],
 :host(:not(:disabled)[variant='contained']) [part='base'] {
-    &:focus,
-    &:active {
+    &:focus-visible {
         box-shadow: 0 0 0 rem(4px) var-get($contained-theme, 'shadow-color');
     }
 }
 
 :host(:not([disabled])[variant='fab']) [part='base'],
 :host(:not(:disabled)[variant='fab']) [part='base'] {
-    &:focus,
-    &:active {
+    &:focus-visible {
         box-shadow: 0 0 0 rem(4px) var-get($fab-theme, 'shadow-color');
     }
 }
--- a/src/components/button/themes/button/shared/button.common.scss
+++ b/src/components/button/themes/button/shared/button.common.scss
@@ -10,16 +10,19 @@ $outlined-theme: $material-outlined;
 :host([variant='flat']) [part='base'] {
     border-radius: var-get($flat-theme, 'border-radius');
     height: var-get($flat-theme, 'size');
+    border: rem(1px) solid var-get($flat-theme, 'border-color');
 }
 
 :host([variant='outlined']) [part='base'] {
     border-radius: var-get($outlined-theme, 'border-radius');
     height: var-get($outlined-theme, 'size');
+    border: rem(1px) solid var-get($outlined-theme, 'border-color');
 }
 
 :host([variant='contained']) [part='base'] {
     border-radius: var-get($contained-theme, 'border-radius');
     height: var-get($contained-theme, 'size');
+    border: rem(1px) solid var-get($contained-theme, 'border-color');
 }
 
 :host([variant='fab']) [part='base'] {
@@ -213,15 +216,6 @@ $outlined-theme: $material-outlined;
 
         ::slotted(igc-icon) {
             color: var-get($fab-theme, 'icon-color-hover');
-        }
-    }
-
-    &:focus {
-        color: var-get($fab-theme, 'focus-foreground');
-        background: var-get($fab-theme, 'focus-background');
-
-        ::slotted(igc-icon) {
-            color: var-get($fab-theme, 'icon-color');
         }
     }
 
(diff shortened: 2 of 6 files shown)


imincheva:
**Button Group**
- Material: Min-width of the button group button should be 50px and Icons shouldn't scale
- Fluent: Min-width of the button group button should be 50px and Icons shouldn't scale
- Bootstrap: Min-width of the button group button should be 42px and Icons shouldn't scale

imincheva:
**Button Group**

- Material: Min-width of the button group button should be 50px and Icons shouldn't scale
- Fluent: Min-width of the button group button should be 50px and Icons shouldn't scale
- Bootstrap: Min-width of the button group button should be 42px and Icons shouldn't scale

SisIvanova:
> @SisIvanova
> 
> Does this PR require the addition of new properties to the button components? Can this be accomplished using only :focus and :focus-visible, or is there something I'm overlooking?

We need this new property to remove the focus styles after clicking on the button which cannot be achieved only using :focus and :focus-visible. The same approach is used in the switch/checkbox/radio components.

rkaraivanov:
> > @SisIvanova
> > Does this PR require the addition of new properties to the button components? Can this be accomplished using only :focus and :focus-visible, or is there something I'm overlooking?
> 
> We need this new property to remove the focus styles after clicking on the button which cannot be achieved only using :focus and :focus-visible. The same approach is used in the switch/checkbox/radio components.

I still believe it's feasible to attain the same functionality using only CSS pseudo-selectors, unless my understanding of the specification is incorrect. The introduction of :focus-visible, which allows different styling based on "browser heuristics" — typically the distinction between pointer and keyboard-driven focus — supports this notion. 

In my experimentation with the toggle button, it appears to exhibit identical behavior after removing the extra JavaScript properties, the **focused** DOM element part, and its references in the Fluent style file, along with implementing the subsequent modifications to the Fluent styles:

```scss
:host {
    [part='toggle'] {
        min-height: $fluent-flat-btn-size;
        border-width: $group-item-border-thickness;
    }

    [part='toggle']:focus {
        background: var-get($theme, 'item-background');
    }

    [part='toggle']:focus-visible {
        background: var-get($theme, 'item-background');

        &::after {
            content: '';
            position: absolute;
            inset-block-start: $outline-btn-indent;
            inset-inline-start: $outline-btn-indent;
            pointer-events: none;
            width: calc(100% - (#{$outline-btn-indent} * 2));
            height: calc(100% - (#{$outline-btn-indent} * 2));
            box-shadow: 0 0 0 rem(1px) var-get($theme, 'item-focused-border-color');
        }
    }
}
```

As I'm not well-versed in CSS, I'm uncertain about the ease, scalability, and applicability of this approach to our component library, so I will limit my comments to the JavaScript aspects. We should consider if it's possible to simplify the implementation of these behaviors without relying on additional JavaScript code.

@simeonoff

src/components/button-group/toggle-button.ts
```diff
@@ -35,6 +36,9 @@ export default class IgcToggleButtonComponent extends LitElement {
   @query('[part="toggle"]', true)
   private _nativeButton!: HTMLButtonElement;
 
+  @state()
+  public focused = false;
```
rkaraivanov, line 40:
```suggestion
  @state()
  protected focused = false;
```

src/components/button-group/button-group.ts
```diff
@@ -171,6 +171,7 @@ export default class IgcButtonGroupComponent extends EventEmitterMixin<
     );
 
     if (button) {
+      button.focused = false;
```
rkaraivanov, line 174:
```suggestion
      Object.assign(button, { focused: false });
```

simeonoff:
> > > @SisIvanova
> > > Does this PR require the addition of new properties to the button components? Can this be accomplished using only :focus and :focus-visible, or is there something I'm overlooking?
> > 
> > 
> > We need this new property to remove the focus styles after clicking on the button which cannot be achieved only using :focus and :focus-visible. The same approach is used in the switch/checkbox/radio components.
> 
> I still believe it's feasible to attain the same functionality using only CSS pseudo-selectors, unless my understanding of the specification is incorrect. The introduction of :focus-visible, which allows different styling based on "browser heuristics" — typically the distinction between pointer and keyboard-driven focus — supports this notion.
> 
> In my experimentation with the toggle button, it appears to exhibit identical behavior after removing the extra JavaScript properties, the **focused** DOM element part, and its references in the Fluent style file, along with implementing the subsequent modifications to the Fluent styles:
> 
> ```css-scss
> :host {
>     [part='toggle'] {
>         min-height: $fluent-flat-btn-size;
>         border-width: $group-item-border-thickness;
>     }
> 
>     [part='toggle']:focus {
>         background: var-get($theme, 'item-background');
>     }
> 
>     [part='toggle']:focus-visible {
>         background: var-get($theme, 'item-background');
> 
>         &::after {
>             content: '';
>             position: absolute;
>             inset-block-start: $outline-btn-indent;
>             inset-inline-start: $outline-btn-indent;
>             pointer-events: none;
>             width: calc(100% - (#{$outline-btn-indent} * 2));
>             height: calc(100% - (#{$outline-btn-indent} * 2));
>             box-shadow: 0 0 0 rem(1px) var-get($theme, 'item-focused-border-color');
>         }
>     }
> }
> ```
> 
> As I'm not well-versed in CSS, I'm uncertain about the ease, scalability, and applicability of this approach to our component library, so I will limit my comments to the JavaScript aspects. We should consider if it's possible to simplify the implementation of these behaviors without relying on additional JavaScript code.
> 
> @simeonoff

Yeah, you are correct in your assessment and we are aware of the focus-visible pseudo selector and we're using it extensively wherever we can. I believe there were some issues that couldn't be overcome by simply using it in the case of the button, but Silvia can expand on that.

rkaraivanov (changes requested):
Avoid exposing additional public APIs for internal states when possible. This approach should accomplish the desired behavior without introducing a new public property.

rkaraivanov (approved):

AnjiManova:
**Button Group**

Material, Fluent & Bootstrap:
- The active state triggered by the keyboard is not working as in the Angular repo for some reason

AnjiManova:
From our side, everything looks fine, and all comments for Material, Fluent and Bootstrap are implemented.

andiesm813:
> **Button Group**
> 
> Material & Fluent:
> 
> * right and left paddings are 16 for all sizes, they should be 8, 12 and 16 (small, medium and large) ✅ implemented
> 
> Bootstrap:
> 
> * right and left paddings are 8, 10 and 12 should be, 8, 12 and 16 (small, medium and large) ✅ implemented
> 
> Indigo:
> 
> * for medium the paddings are correct @andiesm813, could you please provide/validate the small and large size paddings

**INDIGO THEME**

I will just list the correct **left and right** paddings here:

- Small: 6px
- Medium: 8px ✅ 
- Large: 10px

Like @AnjiManova said, the medium size is correctly implemented. The other 2 need to be fixed.

andiesm813:
The comments for Indigo Theme were implemented. So looks good to me!  ✅✅✅

didimmova (approved):

simeonoff (approved):

PR #1134, opened 2024-04-05
Themes: Update Indigo Themes

simeonoff:


SisIvanova (dismissed):

SisIvanova (dismissed):

desig9stein (approved):

AnjiManova:
**Text Area**

- Based on the Input component in the Text Area in the Invalid state, the helper text (hint) shouldn't have color coding. Applicable for Light and Dark mode and Warning and Correct state as well.
image

There is an additional comment related to the Text Area behaviour of saving the filled text and switching to Disabled, Read-only, Invalid, and so on, but we will create a separate issue for this, as the front-end developers recommended.

I saw nothing else as a discrepancy between the UI kit and the implementation.

sbayreva:
**Slider**

1. The track should have 4px radius
2. The tickmarks should start right under the end of the track. 
Screenshot 2024-05-15 at 14 29 34

3. In Figma the the "Discrete Regions" start from 0 
Screenshot 2024-05-15 at 14 33 03
4. The focus border of the thumb should be 3px
5. Тhe triangle of the label tooltip should have 12px height
@andiesm813  please, check the comments and also, please check the behaviour of the Hover and Focus states - whether each element should have these states separately or when on Hover or Focused the whole component should be in that state.

imincheva:
**Tree:**
- The left and right paddings should be fixed to 16px, 12px and 8px for large, medium and small respectively.
- The indentation should be 24px for all display densities.
- Might be irrelevant, but in the UI kits we have a specific stylisation for the hyperlinks, which I believe isn't implemented. Same for the custom chevron, don't know what the situation is with that.
- Тhis is probably some bug, but when I hover the first item, it doesn't inherit the correct color for the foreground in both light and dark mode. In the rest of the items this problem isn't apparent.

andiesm813:
> **Slider**
> 
> 1. The track should have 4px radius
> 2. The tickmarks should start right under the end of the track.
> 
> Screenshot 2024-05-15 at 14 29 34
> 3.Тhe "Discrete Regions" should start from 0 as in the Handoff
> 
> Screenshot 2024-05-15 at 17 11 34
> 4. The focus border of the thumb should be 3px
> 5. Тhe triangle of the label tooltip should have 12px height
>    @andiesm813  please, check the comments and also, please check the behaviour of the Hover and Focus states - whether each element should have these states separately or when on Hover or Focused the whole component should be in that state.

@sbayreva 
All your comments (1-4) are good. About #1 (radius of the track), it was also not implemented in Angular (from what i'm seeing).

About the hover and focus states, the current behavior in WC is the same as in the Angular component. Track and track base do hover together, but the thumb hovers separately, only when hovered on the thumb. So IMO it's all good as implemented now.

PR #1146, opened 2024-04-10
fix(checkbox, switch): participate in form submission when initially checked

### LINKED ISSUE #1144: Checkbox/Switch do not participate in form submission if initially checked
## Description  
What the title of the issue states.

## Steps to reproduce  

1. Create a form with a checkbox/switch initially checked.
2. Submit the form.

## Result  
The key/value pair of the component is not in the form data.

## Expected result  
The key/value pair of the component must be in the form data.

rkaraivanov:
- Dialog - `igcClosed` fired before the component was actually closed/hidden.

Closes #1144


src/components/dialog/dialog.ts
```diff
@@ -206,10 +206,10 @@ export default class IgcDialogComponent extends EventEmitterMixin<
     if (await this.toggleAnimation('close')) {
       this.open = false;
       this.animating = false;
-      this.emitEvent('igcClosed');
     }
 
     await this.updateComplete;
+    this.emitEvent('igcClosed');
```
damyanpetev, line 212: That `if (await this.toggleAnimation('close'))` is a bit odd, since the logic as it reads can potentially not close the dialog (set open to false), though that doesn't seem possible with the current implementation. After discussion, leaving for a separate PR.

damyanpetev (approved):

PR #1157, opened 2024-04-12
fix(tabs, blazor): Wrapper container is undefined

rkaraivanov:


damyanpetev (approved):
Should test eventually if the query cache is really needed, otherwise LGTM

PR #1168, opened 2024-04-18
feat(button-group): clear selection via selectedItems

### LINKED ISSUE #1041: IgcButtonGroup: Setting the variable bound to the selected prop to false doesn't update the UI of the button group
Consider the following sample:

webcomponents button group reset

The idea of the use case is that there would exist a "Reset button" in a given application, whose click handler sets the variable bound to the `selected` property of the toggle buttons in a button group to false.

However, in the UI the toggle button continues appearing as selected even though the click handler of the separate "Reset" button sets the variable to false.

Here is a small sample app with a button group, consisting of two toggle buttons and two separate buttons - one should set the selection to the second of the toggle buttons, and the other should reset the selection:
https://codesandbox.io/p/sandbox/reset-selected-sample-rsxq8k?file=%2Findex.html%3A20%2C116

It would be useful to be able to have the styling of the toggle buttons in a button group be updated to selected or de-selected, when setting the variable bound to the `selected` property of the toggle button.

RivaIvanova:
Closes #1041

### DIFF
--- a/src/components/button-group/button-group.ts
+++ b/src/components/button-group/button-group.ts
@@ -109,10 +109,8 @@ export default class IgcButtonGroupComponent extends EventEmitterMixin<
   }
 
   public set selectedItems(values: string[]) {
-    if (values) {
-      this._selectedItems = new Set(values);
-      this.setSelection(this._selectedItems);
-    }
+    this._selectedItems = new Set(Array.isArray(values) ? values : []);
+    this.setSelection(this._selectedItems);
   }
 
   @watch('disabled', { waitUntilFirstUpdate: true })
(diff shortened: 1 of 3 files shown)


src/components/button-group/button-group.ts
```diff
@@ -109,7 +109,7 @@ export default class IgcButtonGroupComponent extends EventEmitterMixin<
   }
 
   public set selectedItems(values: string[]) {
-    if (values && values.length) {
+    if (values) {
       this._selectedItems = new Set(values);
       this.setSelection(this._selectedItems);
     }
```
rkaraivanov, line 115:
```suggestion
    this._selectedItems = new Set(Array.isArray(values) ? values : []);
    this.setSelection(this._selectedItems);
```
This would also work when assigning falsy values, i.e. `null/undefined`.
You can add a test case for that as well.

rkaraivanov (changes requested):
LGTM with one nitpick.

Also, please add a changelog entry with the feature.

rkaraivanov (approved):

PR #1178, opened 2024-04-23
feat(input,textarea): Added validate-only

rkaraivanov:
- [x] Agree on the naming of the property.
- [x] Update related specifications
- [x] Add CHANGELOG entry


src/components/input/input.ts
```diff
@@ -247,6 +247,15 @@ export default class IgcInputComponent extends IgcInputBaseComponent {
   @property()
   public autocomplete!: string;
 
+  /**
+   * Consider whether to permit user input to exceed the `maxLength` when it is specified for a string type input,
```
damyanpetev, line 251: Kind of thinking the wording here can be more related to the option, because while it would allow to exceed the limit, it still validates or only validates if you will :)
So considering something in the lines of:
```suggestion
   * Allows validation constraints to be checked without restricting user input. Applies to `maxLength` for a string type input,
```
and so on.
Does that sound clear enough?

rkaraivanov: What do you think of this one?

> Enables validation rules to be evaluated without restricting user input. This applies to the `maxLength` property for string-type inputs or allows spin buttons to exceed the predefined `min/max` limits for number-type inputs.

damyanpetev (approved):

PR #1182, opened 2024-04-25
fix(date-time input): Label in Material with read-only

### LINKED ISSUE #1166: IgcDateTimeInput: label is not positioned correctly when readOnly=true
## Description  
The label in material theme is not positioned correctly when readOnly=true for the `IgcDateTimeInput` (used by the `IgcDatePicker` as well).

## Steps to reproduce  

1. Set readOnly to 'true'

## Result  
See attachment.

## Expected result  
As the `IgcInput` with a placeholder:
image

## Attachments  
[Attach a sample if available, and screenshots, if applicable.
](https://github.com/IgniteUI/igniteui-webcomponents/assets/127843837/978265b2-ae39-46ed-90d3-b80bfc35bc4c)

rkaraivanov:
* Additionally cleaned up some low-hanging code for easier reading.

Closes #1166

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Dropdown `positionStrategy` property. The dropdown now uses the Popover API to render its container in the top layer of the browser viewport,
   making the property obsolete.
 
+### Fixed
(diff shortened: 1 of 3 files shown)


rkaraivanov:
@damyanpetev 

I've bumped the ES lib version in the tsconfig since we should really target 2022 baseline for browsers.
Check [this](https://caniuse.com/mdn-javascript_builtins_array_findlast) and [this](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility).

damyanpetev (approved):

PR #1196, opened 2024-05-13
Themes: Add updated Indigo themes for Navbar, Navdrawer, and Switch

simeonoff:
- [x] Navbar
- [x] Navigation Drawer
- [x] Switch


AnjiManova:
**Switch**

Dark Mode:
- The checked switch hover should be primary.300 (--track-on-hover-color & --border-on-hover-color)
- The checked switch focused & hover should be primary.300 (--track-on-hover-color & --border-on-hover-color)

AnjiManova:
**Navbar**

all good

imincheva:
**Nav Drawer:**
- The min-height of the header content is missing. It should be 16px, in order to achieve the 32px height of the header.
- The width of the Nav Drawer should be 48px. I also noticed that in Indigo the items only have paddings, while in the rest of the themes the items have paddings and margins (IMO this should be treated as an inconsistency). In Indigo these margins are applied on <div part="main">

desig9stein (approved):

PR #1198, opened 2024-05-14
refactor: Removed blazor decorators

rkaraivanov:
Removed blazor alaternateName and blazorTwoWayBind and migrated any members decorated with those
to trivia comments.


src/components/dropdown/dropdown.ts
```diff
@@ -368,8 +367,8 @@ export default class IgcDropdownComponent extends EventEmitterMixin<
   public navigateTo(value: string): IgcDropdownItemComponent | null;
   /** Navigates to the item at the specified index. If it exists, returns the found item, otherwise - null. */
   public navigateTo(index: number): IgcDropdownItemComponent | null;
+  /* blazorSuppress */
```
damyanpetev, line 370: Because it's no longer a decorator that resolves on each overload rather leading trivia that's separate, at least with the current impl that doesn't distinguish the main method you'll need to add `blazorSuppress` to the two overloads as well

src/components/select/select.ts
```diff
@@ -563,8 +561,8 @@ export default class IgcSelectComponent extends FormAssociatedRequiredMixin(
   public navigateTo(value: string): IgcSelectItemComponent | null;
   /** Navigates to the item at the specified index. If it exists, returns the found item, otherwise - null. */
   public navigateTo(index: number): IgcSelectItemComponent | null;
+  /* blazorSuppress */
```
damyanpetev, line 564: Same as the dropdown

damyanpetev (approved):

PR #1204, opened 2024-05-16
Themes: Refactor Checkbox, Switch, and Radio

simeonoff:


sbayreva:
**Fluent**

CHECKBOX:
1. For Empty Invalid checkbox when on Hover the tick should be  Error 500 
2. Something I've noticed- there is a color style called “ disabled-indeterminate-color” that uses secondary 100, and it is not used with the actual disabled indeterminate state here which is good because the right color is grays.400. 

RADIO BUTTON:
All good

SWITCH:
All good

imincheva:
**Bootstrap:**
- Checkbox: by design
- Radio: by design
- Switch: by design

AnjiManova:
**Material**

Checkbox
- Light Mode: invalid/indeterminate - in the hover state, the ripple (:before) should be error.500 with 0.06% opacity

AnjiManova:
**Material**

Radio Group and Radio

- I'm not sure if this is a problem, but I cannot trigger the Focused state with focused in the inspect mode like in the other themes. In Material, the focused state is visible on focus-within
- After selecting a radio item, it stays focused and this might be related to the comment above^^

AnjiManova:
**Material**

Switch

Light Mode
- the ripple should be grays.600 with 6% opacity on hover
- the ripple should be grays.600 with 12% opacity when is focused

sbayreva:
> **Fluent**
> 
> CHECKBOX:
> 
> 1. For Empty Invalid checkbox when on Hover the tick should be  Error 500
> 2. Something I've noticed- there is a color style called “ disabled-indeterminate-color” that uses secondary 100, and it is not used with the actual disabled indeterminate state here which is good because the right color is grays.400.
> 
> RADIO BUTTON: All good
> 
> SWITCH: All good



> **Fluent**
> 
> CHECKBOX:
> 
> 1. For Empty Invalid checkbox when on Hover the tick should be  Error 500
> 2. Something I've noticed- there is a color style called “ disabled-indeterminate-color” that uses secondary 100, and it is not used with the actual disabled indeterminate state here which is good because the right color is grays.400.
> 
> RADIO BUTTON: All good
> 
> SWITCH: All good

Fixed!

desig9stein (dismissed):

didimmova (approved):

PR #1221, opened 2024-05-28
feat(slider): update indigo slider

### LINKED ISSUE #1213: Slider: Theme Issues
1. The track should have 4px radius
2. The tickmarks should start right under the end of the track. 
Screenshot 2024-05-15 at 14 29 34

3.Тhe "Discrete Regions" should start from 0 as in the Handoff

Screenshot 2024-05-15 at 17 11 34

4. The focus border of the thumb should be 3px

5. Тhe triangle of the label tooltip should have 12px height
@andiesm813  please, check the comments and also, please check the behaviour of the Hover and Focus states - whether each element should have these states separately or when on Hover or Focused the whole component should be in that state.

_Originally posted by @sbayreva in https://github.com/IgniteUI/igniteui-webcomponents/issues/1134#issuecomment-2112293500_

Comment from Andie:

@sbayreva
All your comments (1-4) are good. About https://github.com/IgniteUI/igniteui-webcomponents/issues/1 (radius of the track), it was also not implemented in Angular (from what i'm seeing).

About the hover and focus states, the current behavior in WC is the same as in the Angular component. Track and track base do hover together, but the thumb hovers separately, only when hovered on the thumb. So IMO it's all good as implemented now.

didimmova:
Closes #1213


sbayreva:
1. The space between the tick marks and the base track should be 4px
2. Not sure whether this is a problem but when inspected the marking rectangular is not in the center of the thumb.
Screenshot 2024-05-29 at 14 23 00

@andiesm813 Please, confirm and add if there is anything else.

sbayreva:
> 1. The space between the tick marks and the base track should be 4px
> 2. Not sure whether this is a problem but when inspected the marking rectangular is not in the center of the thumb.
> 
> Screenshot 2024-05-29 at 14 23 00
> 3. The space between tickmarks and labels should be 4px
> 
> @andiesm813 Please, confirm and add if there is anything else.

Verified by my side, waiting for @andiesm813 approval.

SisIvanova (approved):

andiesm813:
All looks good to me! ✅

PR #1232, opened 2024-06-05
fix(IgcCombo): Add a tag that generates generic type for blazor.

MayaKirova:


src/components/input/input.ts
```diff
@@ -47,6 +47,8 @@ export default class IgcInputComponent extends IgcInputBaseComponent {
   protected _value = '';
 
   /* @tsTwoWayProperty(true, "igcChange", "detail", false) */
+  /* blazorGenericType */
+  /* blazorAlternateType: object */
```
damyanpetev, line 51: Might end up moving this to the input base because that's the one we need, but the generated models by this are already merged, so it'll me a separate change.

damyanpetev (approved):

PR #1254, opened 2024-06-13
feat: Render validation messages for Form associated custom elements

### LINKED ISSUE #1043: IgcDateTimeInput - missing invalid event or invalid prompt display
IgcDateTimeInput is missing an invalid event or invalid prompt display to inform the user to enter a valid value. The placeholder property is suitable but not desirable in all scenarios.

### LINKED ISSUE #1195: Enhancing Web Components Form Validation
Here is a sample that demonstrate the current way to implement Web Components form with validation: [Web Components Validation](https://github.com/Marina-L-Stoyanova/FormsBuilder/tree/main/All%20components%20samples/WebComponents%20validation)

We should consider:
1. In-Component Validation:
     > Implement validation logic directly within the Web Components that can be used in a Form(igc-input, igc-textarea, igc-select and ect.). This approach encapsulates validation rules and keeps them self-contained.
     > Validate input data as users interact with the form elements (e.g., on blur or input change events).
2. Displaying Validation Messages:
   > Show validation messages when input is invalid. 
3. Slot Considerations:
    > Evaluate whether to use existing slots (like helper-text) or create a dedicated slot (e.g., validation-message) for displaying validation feedback. (Using a separate slot allows flexibility in styling and positioning validation messages.)

rkaraivanov:
Closes #1195 
Closes #1043 

Specification [here](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Form-associated-elements-validation-slots)


imincheva:
**All themes**
- Icon should be error, not cancel
- Line height of the validation slot is set to 1. It should come from the respective text style used for the validation message text.

**Material**
- 16px right padding is missing on the validation slot

imincheva:
Validation message on the Text Area in Bootstrap has 16px left padding, which should be removed.

imincheva:
Validation message on the Date picker in Bootstrap has 8px left padding, which should be removed, but I see this is applied on the hint message as well, so not sure whether it should be fixed in this PR or logged as a separate issue (This padding isn't applied in the rest of the input controls like Input, Select, Combo etc) The gap between input field and hint/validation message is also missing - it should be 4px.

src/components/common/mixins/forms/types.ts
```diff
@@ -78,12 +81,12 @@ declare class BaseFormAssociatedElement {
...
    * if the element has completed the first update cycle or it has been interacted with by the user.
    */
-  protected _validate(message?: string): void;
+  protected _validate(message?: string, isUserSet?: boolean): void;
 
   /**
    * Executes the component's validators and updates the internal validity state.
    */
-  protected _updateValidity(message?: string): void;
+  protected _updateValidity(message?: string, isUserSet?: boolean): void;
```
damyanpetev, line 89: Are those some sort of leftover? Don't see any code calling in with the second param nor is it implemented since `BaseFormAssociated` doesn't actually have a connection to this type.

src/components/checkbox/checkbox.spec.ts
```diff
@@ -382,4 +387,41 @@ describe('Checkbox', () => {
...
+
+    it('renders invalid slot', async () => {
+      await createFixture(html`
+        <igc-checkbox required>
+          <div slot="invalid"></div>
+        </igc-checkbox>
+      `);
+
+      await checkValidationSlots(element, 'invalid');
+    });
```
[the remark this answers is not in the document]
damyanpetev: Pretty much true for the rest of the components running this set of checks it seems.

src/components/input/input-base.ts
```diff
@@ -107,6 +105,10 @@ export abstract class IgcInputBaseComponent extends FormAssociatedRequiredMixin(
 
   protected abstract renderInput(): TemplateResult;
 
+  protected renderValidatorContainer(): TemplateResult {
+    return IgcValidationContainerComponent.create(this);
```
damyanpetev, line 109: Hm, noticed doesn't specify an id like the combo and doesn't really use `aria-describedby` either. Somewhat assumed the re-slotting of `helper-text` from other components would at least solve this here, but seems like a nope. I know, more like a separate issue - feel free to log.

simeonoff (approved):

damyanpetev (approved):

PR #1261, opened 2024-06-18
fix(combo): resolve #1229 #1238 #1260

### LINKED ISSUE #1229: [Bug]: igc-combo loses the last selected item with the keyboard

Combo

### Description

When I select items with the keyboard, the item selected last is lost on submission.

### Reproduction

1. Go to https://igniteui.github.io/igniteui-webcomponents/?path=/story/combo--basic
2. Focus the combo
3. Go down the items with the keyboard
4. Select an item that is currently not selected by pressing the spacebar
5. Press enter

### Workaround

Not using the keyboard to select.

### Is this a regression?

### Affected versions

4.9.0

Browser: Edge Dev

### LINKED ISSUE #1238: [Bug]: Selected value not displayed in input after reselecting the same value in igc-combo

Combo

### Description

We have a single-select igc-combo with a pre-selected value. When we focus the combo input, highlight the existing value, and enter the first letter of that value to filter the drop-down list, we can see the same value in the filtered list. However, upon reselecting the same highlighted value, the input field does not display the selected value; instead, it only shows the letter that was typed.

### Reproduction

1. Open the igc-combo demo for single selection at [this link](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/combo/single-selection#web-components-single-selection-example).
2. Select a value from the combo (e.g., "London").
3. Focus the combo input field, and the value "London" should be highlighted.
4. Type the first letter of the selected value (e.g., 'L').
5. The combo's drop-down list should filter to show values starting with 'L', including "London".
6. Reselect "London" from the filtered list.

### Expected Result
The input field should display the full selected value ("London") after reselecting it.

### Actual Result
The input field only displays the letter that was typed ('L') instead of the full selected value ("London").

### Workaround

No

### Is this a regression?

### Affected versions

Latest

Browsers: All

### LINKED ISSUE #1260: [Bug]: Combo retains search term when toggling filtering

Combo

### Description

When toggling the filtering by using the `disableFiltering` option on the combo, the combo search field persists the term while it shouldn't. This also makes the combo persist the filtered stated if the combo is initially filtered by a term and the disableFiltering option is then set to true, rendering the combo unusable for wider selection.

Here's a short clip:

https://github.com/IgniteUI/igniteui-webcomponents/assets/4118303/d3f28f8d-d3d5-4e82-b4c4-9e559988dd46

### Reproduction

1. Open the combo sample
2. Filter the list of items
3. Disable filtering by setting disableFiltering to true
4. Open the combo and see that the filtering is not reset

### Workaround

I haven't found a workaround for this problem.

### Is this a regression?

### Affected versions

Failing in 4.8.0

Browser: All

simeonoff:
Closes #1229 
Closes #1238 
Closes #1260


rkaraivanov (changes requested):
LGTM

Can you add entries in the changelog with the fixed issues?

PR #1262, opened 2024-06-18
feat(radio-group): Set radio properties through the group

### LINKED ISSUE #315: Cannot bind radio group to a value
## Description  
There is no way to bind radio group component to a value.

## Steps to reproduce  

1. Add radio group.
2. Add three radios, e.g. male, female and other.
3. Add a component property, e.g. gender.
4. Try to bind radio group to the gender property.

## Result  
There is no way to bind the radio group.

## Expected result  
It should be possible to bind radio group to a component's property. Also if the value of any of the radios in the group equals the value of the bound property this radio should be checked.

rkaraivanov:
Added `name` and `value` properties on the igc-radio-group. Developers can now control the checked state of the underlying igc-radio components through the enclosing group.

Closes #315

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - Banner component [#1174](https://github.com/IgniteUI/igniteui-webcomponents/issues/1174)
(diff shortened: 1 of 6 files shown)


wnvko:
LGTM
Hopefully no one will set the radio group name to 0 of `false`.

rkaraivanov:
> LGTM Hopefully no one will set the radio group name to 0 of `false`.

This shouldn't be an issue in normal scenarios. If it is bound as an attribute both 0 and false are coerced to truthy string values. If you try to set them through a property assignment on a **string** property that is just a developer being sloppy.

PR #1295, opened 2024-07-09
feat(icon-registry): add broadcast channel to sync icons.

MayaKirova:
Goes with: https://github.com/IgniteUI/igniteui-angular/pull/14498


src/components/icon/icon.registry.ts
```diff
@@ -11,12 +11,24 @@ import type {
...
+  SyncState = 0,
+  RegisterIcon = 1,
+  UpdateIconReference = 2,
+}
+
+interface BroadcastIconsChangeMessage {
+  actionType: ActionType;
+  collections?: Map<string, Map<string, SvgIcon>>;
+  references?: Map<string, Map<string, IconMeta>>;
+}
```
rkaraivanov, line 24: Can you move these into the _registry/typest.ts_ module?

src/components/icon/icon.registry.ts
```diff
@@ -25,6 +37,21 @@ export class IconsRegistry {
...
+        const userSetCollection: Map<string, Map<string, SvgIcon>> = this.getUserSetCollection();
+        const refs: Map<string, Map<string, IconMeta>> = this.getUserSetRefs();
+        const message: BroadcastIconsChangeMessage = {
+          actionType: ActionType.SyncState,
+          collections: userSetCollection,
+          references: refs
+        };
+        this.iconBroadcastChannel.postMessage(message);
+      }
+    };
```
rkaraivanov, line 54: Maybe we can move the whole logic/state around the broadcast channel into a different module/class?

rkaraivanov (changes requested):

src/components/icon/icon-state.broadcast.ts
```diff
@@ -0,0 +1,89 @@
...
+  private iconBroadcastChannel: BroadcastChannel;
+  constructor(
+    collections: Collection<string, Map<string, SvgIcon>>,
+    refsCollection: Collection<string, Map<string, IconMeta>>
+  ) {
+    this.parser = new SvgIconParser();
+    this.iconBroadcastChannel = new BroadcastChannel('ignite-ui-icon-channel');
+    this.iconBroadcastChannel.onmessage = (event) => {
+      const message = event.data as BroadcastIconsChangeMessage;
+      if (message.actionType === ActionType.SyncState) {
```
rkaraivanov, line 22: I suppose this work in progress. Don't we have to update the state of the icon registry when we receive `RegisterIcon`/`UpdateIconReference`. 
I guess we also have to update the state in the initial sync as well since this registry can be created after the angular icon service is already up and having a few custom icons.

src/components/icon/registry/types.ts
```diff
@@ -31,3 +31,15 @@ export interface IconMeta {
...
+  SyncState = 0,
+  RegisterIcon = 1,
+  UpdateIconReference = 2,
+}
+
+export interface BroadcastIconsChangeMessage {
+  actionType: ActionType;
+  collections?: Map<string, Map<string, SvgIcon>>;
+  references?: Map<string, Map<string, IconMeta>>;
+}
```
rkaraivanov, line 45: Maybe move this above the publicly exported types so future us don't get confused why they are not exported from the main entry point of the package

src/components/icon/icon-state.broadcast.ts
```diff
@@ -0,0 +1,89 @@
...
+export class IconsStateBroadcast {
+  private parser: SvgIconParser;
+  private iconBroadcastChannel: BroadcastChannel;
+  constructor(
+    collections: Collection<string, Map<string, SvgIcon>>,
+    refsCollection: Collection<string, Map<string, IconMeta>>
+  ) {
+    this.parser = new SvgIconParser();
+    this.iconBroadcastChannel = new BroadcastChannel('ignite-ui-icon-channel');
+    this.iconBroadcastChannel.onmessage = (event) => {
```
rkaraivanov, line 20: Nitpick: I would prefer this to be `addEventListener('message', handler)`, and make the handler a separate class method.

src/components/icon/icon-state.broadcast.ts
```diff
@@ -0,0 +1,89 @@
...
+            userSetIcons.set(collectionKey, new Map<string, SvgIcon>());
+            icons = userSetIcons.get(collectionKey);
+          }
+          if (icons) {
+            icons.set(iconKey, this.parser.parse(val));
+          }
+        }
+      }
+    }
+    return userSetIcons;
```
rkaraivanov, line 76: Ok, this really ramps up the cognitive effort when reading the code. Can you somewhat "clean" it up?

rkaraivanov (changes requested):

src/components/icon/icon-state.broadcast.ts
MayaKirova: Do we actually need that? There's no public API exposed from angular element's side that would allow us to update the icons. So how would we change them from the angular icon service?

rkaraivanov: Fair point. I was completely absent-minded that the angular elements stuff relies on the WC public API 💀

Then in that case can you update the tests to only check for the initial sync status on another peer. Maybe add a couple around the icon refs event

rkaraivanov (approved):

damyanpetev (dismissed):

damyanpetev (approved):

PR #1303, opened 2024-07-12
refactor(checkbox, switch): `igcChange` event type

rkaraivanov:
breaking change

### DIFF
--- a/src/components/checkbox/checkbox-base.ts
+++ b/src/components/checkbox/checkbox-base.ts
@@ -14,7 +14,7 @@ import {
 } from '../common/validators.js';
 
 export interface IgcCheckboxEventMap {
-  igcChange: CustomEvent<boolean>;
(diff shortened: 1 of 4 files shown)


src/components/checkbox/checkbox-base.ts
```diff
@@ -14,7 +14,7 @@ import {
 } from '../common/validators.js';
 
 export interface IgcCheckboxEventMap {
-  igcChange: CustomEvent<boolean>;
+  igcChange: CustomEvent<{ checked: boolean; value: string | undefined }>;
```
damyanpetev, line 17: Any chance we can have a type with name for these event args?
I know this works fine, but for the analyzer and other platforms as a general rule rule we should really try to avoid anonymous types - it either has to come up with a name regardless (which doesn't always account for shared types) or in this case fails to gen in the current impl.

PR #1313, opened 2024-07-17
feat(icons): add icon references

### LINKED ISSUE #1304: Icons: Use icons by reference in the component templates
- [x] Research
- [x] Specification
- [x] API Documentation
- [ ] Design implementation

simeonoff:
Closes #1304


src/components/icon/icon.registry.ts
```diff
@@ -28,44 +43,52 @@ export class IconsRegistry {
...
+      reference.set(alias.name, { ...target });
+    }
+
+    this.notifyAll(alias.name, alias.collection);
   }
 
-  public register(name: string, iconText: string, collection = 'default') {
-    const namespace = this.getOrCreateCollection(collection);
-    namespace[name] = this.parseSVG(iconText);
+  public getIconRef(name: string, collection: string): IconMeta {
```
damyanpetev, line 75: Just as a note: Could consider eventually exposing this to allow user code to interrogate the registry state if needed (think we've did so in Angular already), but can wait for actual usage to ramp up.

src/components/icon/icon.registry.ts
```diff
@@ -28,44 +43,52 @@ export class IconsRegistry {
...
+      for (const { alias, target } of iconReferences) {
+        const external = this.references
+          .get(alias.collection)
+          ?.get(alias.name)?.external;
+
+        this.setIconRef({
+          alias,
+          target: target.get(this.theme) ?? target.get('default')!,
+          overwrite: !external,
+        });
```
damyanpetev, line 59: Doubt it'll be a significant difference atm (there are pretty much 17 default icons in iconReferences), but this could use some optimization to check if the target actually changes to avoid running `notifyAll` for all entries - either here or in general built into `setIconRef`

src/components/rating/rating.ts
```diff
@@ -343,13 +343,13 @@ export default class IgcRatingComponent extends FormAssociatedMixin(
       const { forward, backward } = this.clipSymbol(i, ltr);
       yield html`<igc-rating-symbol exportparts="symbol, full, empty">
         <igc-icon
-          collection="internal"
-          name="star"
+          collection="default"
+          name="star-filled"
```
damyanpetev, line 347: If the pattern is using underscores this should be `star_filled` and same for `star_outlined`, FYI

src/theming/theming-decorator.ts
```diff
@@ -1,38 +1,71 @@
...
+/**
+ * Returns the theming controller for the given element if exposed.
+ */
+export function getThemeController(host: LitElement) {
+  return isControllerExposed(host) ? host[themeSymbol] : undefined;
+}
+
+function isControllerExposed(
+  host: LitElement
+): host is LitElement & { [themeSymbol]: ThemingController } {
```
damyanpetev, line 66: @rkaraivanov Isn't this supposed to return the `ThemeController` interface? Otherwise that interface is goes pretty unused atm.

damyanpetev (approved):
Overall LGTM and I know there will be a follow-up PR to adjust some naming, so I'm going to merge this to allow work on the Elements integration to continue.
Besides some minor comments, @simeonoff a reminder for changelog and perhaps a similar table of icon usage (I know.. 😶) could be useful in source as well.

PR #1321, opened 2024-07-22
Date Time Input: update masked value on focus when value is set

### LINKED ISSUE #1320: [Bug]: Date Time Input does not switch between the input and display formats on focus when value is set

Date-time input, Date picker

### Description

The expectation when focusing the input is that the mask is updated according to the inputFormat, so that the value can be edited as this format requires. However, this is not the case when both `inputFormat` and `displayFormat` properties are set and the input has a value. The mask remains as the `displayFormat`.

### Reproduction

1. Check out the demo under [this](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/date-time-input#input-format) section.
2. Focus the input and enter a value, such as the today date by pressing `Arrow Up`.
3. Blur the input.
4. Click on it to focus again.
5. Observe the mask remaining as the display format.
6. Test the editing by deleting some characters for example.
7. Blur the input.
8. The value is not properly edited and becomes `Feb 1, 2000, 12:00:00 AM`.

dte-issue

### Workaround

no

### Is this a regression?

### Affected versions

4.9.0

any

ddaribo:
Closes #1320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Date-time-input - `minValue`/`mavValue` are removed. Use `min`/`max` instead.
 - Removed size property from components.
 
+### Fixed
+- Date-time input - update masked value according to the input format on focus when value is set [#1320](https://github.com/IgniteUI/igniteui-webcomponents/issues/1320)
+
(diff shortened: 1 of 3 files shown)


rkaraivanov (changes requested):
LGTM

Please add an entry in the changelog

PR #1325, opened 2024-07-23
removal: Remove custom focus/blur events

rkaraivanov:


damyanpetev:
Looks like we'll need to maintain the events in the event maps for analyzer meta, otherwise this seems to be working great.
Even handles the Combo internal switching between main and search inputs without throwing blurs at no extra check cost :)

damyanpetev (approved):

PR #1326, opened 2024-07-23
refactor:  form-associated validators

rkaraivanov:
* Moved apart the component validators into their own files.
* Renamed the internal `validators` property and made it read-only.


src/components/radio/validators.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { Validator } from '../common/validators.js';
+import type IgcRadioComponent from './radio.js';
+
+export const radioValidators: Validator<IgcRadioComponent>[] = [
+  {
+    key: 'valueMissing',
+    message: messages.required,
+    isValid: (host) => {
+      // @ts-expect-error - protected access
+      const [radios, checked] = [host._radios, host._checkedRadios];
```
damyanpetev, line 11: Oh I thought `getGroup` was actually locked to usage with group parent, but actually _both_ `_radios` and `_checkedRadios` are proxies of the same logic, so this would be _much_ cleaner w/ `getGroup(host)` and you save one run of it since the getters don't cache too.

```suggestion
import type IgcRadioComponent from './radio.js';
import { getGroup } from './utils.js';

export const radioValidators: Validator<IgcRadioComponent>[] = [
  {
    key: 'valueMissing',
    message: messages.required,
    isValid: (host) => {
      const { radios, checked } = getGroup(host);
```

PR #1332, opened 2024-07-26
refactor: Migrate common union types to a type alias

rkaraivanov:


src/components/types.ts
```diff
@@ -0,0 +1,3 @@
+export type Variant = 'primary' | 'info' | 'success' | 'warning' | 'danger';
```
damyanpetev, line 1: There should probably be exported from the public API, no? Even though being string unions, still might be useful to be able to reference those for typing stuff perhaps.

damyanpetev: Also "Variant" and "WeekDays" might be a bit too generic/common/in danger of collisions.
Will get back on that with suggestions

src/components/chip/chip.ts
```diff
@@ -81,7 +82,7 @@ export default class IgcChipComponent extends EventEmitterMixin<
    * @attr
    */
   @property({ reflect: true })
-  public variant!: 'primary' | 'success' | 'danger' | 'warning' | 'info';
+  public variant!: Variant;
```
damyanpetev, line 85: Oh cool, one more thing I noticed, technically this is the other way around: `variant?` instead of `variant!`.
Either that, or the chip has an extra state the badge does not - default , which isn't any of the variants.

src/components/types.ts
ChronosSF: `WeekDays` works better being generic. I wouldn't change it. It is just what it says and can be used separately from the calendar if someone just needs such an enum.

ChronosSF (changes requested):

src/components/types.ts
damyanpetev: I was thinking `CalendarWeekDay` but honestly `WeekDays` (plural) works just as well. From what I see most .NET stuff already in use are using `WeekDay` struct which should not conflict, so we can keep it and it was an existing type already.

As for the variant - either `ColorVariant` since it's both chip & badge, or going of the description in the docs https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/chip#variants
>  chip supports several pre-defined stylistic variants

perhaps `StyleVariant` is more like it and it's reasonably unique and should be semantically correct.

PR #1336, opened 2024-07-30
refactor(icon-service): update icon references

simeonoff:


src/components/icon/icon-references.ts
```diff
@@ -6,6 +6,7 @@ import type {
 
 type Icon = { [key in IconThemeKey]?: IconMeta };
 
+export const iconReferences: IconReference[] = [];
```
rkaraivanov, line 9: Any particular reason to create an empty array and add refs later rather than inline the whole thing?

simeonoff: The only reason is the declaration of individual icons is streamlined and takes less space as source code.

rkaraivanov (approved):

PR #1340, opened 2024-08-02
fix: Programmatic showing of popovers

### LINKED ISSUE #1339: [Bug]: IgcDropdown is not opened on snackbar action button click

Dropdown

### Description

Drop-down is not opened on snackbar action button click.

### Reproduction

1. Run the attached sample
2. Click on the Button
3. Click on the Snackbar action button
[untitled-app (12)_webComponents (1).zip](https://github.com/user-attachments/files/16455138/untitled-app.12._webComponents.1.zip)

### Workaround

`  private openDropDown(){
    requestAnimationFrame(()=>{
      this.dropdown?.toggle(this.btn);
    })
  }`

### Is this a regression?

### Affected versions

4.11

N/A

rkaraivanov:
Skip the root click contoller logic when programmatically opening a popover in response to a click event that is outside the DOM of the popover parent.

Closes #1339


damyanpetev (approved):
Yeah, bit awkward timing issue unfortunately, could use another look in the future but the disable flag should do for now and this fix is kind of important and will do for now :)

PR #1352, opened 2024-08-16
Refactor Tab component

### LINKED ISSUE #1289: Resurrect Tab refactoring PR
[This](https://github.com/IgniteUI/igniteui-webcomponents/pull/627) PR in relation to [implementing tab as a single component](https://github.com/IgniteUI/igniteui-webcomponents/issues/504) has been made more than an year ago.  Please investigate what is the better approach to proceed further - resolve all conflicts and make the branch equal to master or start it from scratch.

MonikaKirkova:
Closes #1289


yradoeva:
**Relevant for all themes:**

1. The right scroll button should go in disabled state, when you reach the end of the tabs (as the left one is disabled when you are in the beginning). Currently the right button stays enabled. 


https://github.com/user-attachments/assets/2367eab1-b5b9-40a5-a706-aefd2061826f

yradoeva:
**Fluent - Light & Dark Mode**
The height of the tall tab item should be 72 px, top and bottom paddings should be 10 px.
Font 16px Open Sans, Helvetica, Arial, sa

**Fluent - Light Mode**
1. **Tab Item - Inactive, Idle** - Icon color should be grays.800 (not grays.700)
Label with suffixprefix
.
Pasted Graphic 8

2. **Tab Item - Active, Hover:** text color and icon color should be grays.900 (not grays.800)
Pasted Graphic 7
.
Pasted Graphic 10

3. **Disabled Scroll Button** - background color should be surface light (as far as I see, it’s transparent)
, background-color



**Fluent - Dark Mode:**
1. **Tab Item - Inactive, Idle** - Icon color should be grays.800 (not grays.700)
Label with suffixprefix

2. **Tab Item - Inactive/Active - Hover:** background color should be grays.50 (not grays.100 0,5)
Inactive
.
--item-hover-background var(

3. **Tab Item - Active, Hover:** text color and icon color should be grays.900 (not grays.800)
Pasted Graphic 17

4. **Disabled Button** - background color should be surface dark (as far as I see it’s transparent) and icon color should be grays.200 (not grays.400)
Pasted Graphic 19
vs. design:
Pasted Graphic 18

yradoeva:
**Bootstrap - Light & Dark Mode**

**Tall Tab Item (Tab item with icon and text)** should have top and bottom paddings of 8px (not 12px)
image

yradoeva:
**Indigo**
**Light & Dark Mode**

1. **Scroll Button width** should be 40px (not 48px)
Color color(srgb 0 423529 0 439216

2. Currently the Active Indicator appears below the tabs bottom border line. If possible, the Active Indicator should be over the bottom border.
Pasted Graphic 24

yradoeva:
**Material Theme - all looks good**

desig9stein:
> **Fluent - Light & Dark Mode**
> The height of the tall tab item should be 72 px, top and bottom paddings should be 10 px.

@yradoeva In the UI kit when there is only text there is a 12px padding, top and bottom, when there is a text + icon or just icon the padding is 10px. I can't make it work that way here, please, pick one for all.

yradoeva:
> > **Fluent - Light & Dark Mode**
> > The height of the tall tab item should be 72 px, top and bottom paddings should be 10 px.
> 
> @yradoeva In the UI kit when there is only text there is a 12px padding, top and bottom, when there is a text + icon or just icon the padding is 10px. I can't make it work that way here, please, pick one for all.

Let's take the padding from the tall tab item (icon+text), top and bottom padding to be 10px. As far as I see, the tab item has a min-height of 2.75 rem and that way the short one (only text) will correspond to the design.

desig9stein:
> 2\. Currently the Active Indicator appears below the tabs bottom border line. If possible, the Active Indicator should be over the bottom border.

Can you confirm this one? Asking, since for me the indicator is on top of the line.

yradoeva:
> > 2. Currently the Active Indicator appears below the tabs bottom border line. If possible, the Active Indicator should be over the bottom border.
> 
> @yradoeva Can you confirm this one? Asking, since for me the indicator is on top of the line.

Well, yes, I see it below the line, it's more visible in dark mode, and when you zoom in. I've set the line to "red" to be more clear. Not to be misunderstood, both lines should overlap with 1px, just the active indicator should be the top layer. For comparison, in Angular it looks as it should be.

https://github.com/user-attachments/assets/a01b1b6e-09bb-40d9-a0c2-a72685d2001d

yradoeva:
**Fluent** - previous comments are resolved, the one thing left to check if possible, is for the Active Tab items to have different text/icon colors when in idle and in hover state. Values by design: 
image

yradoeva:
**Fluent - Dark mode**
The color of the indicator line for the Active Tab item should be primary.300. 
image

src/components/tabs/themes/tabs.base.scss
```diff
@@ -1,116 +1,155 @@
...
+    transition: none;
 }
 
-:host-context([dir='rtl']),
-:host([dir='rtl']) {
-    [part='start-scroll-button'],
-    [part='end-scroll-button'] {
-        transform: scaleX(-1);
+[part='selected-indicator'] {
+    $indicator-bezier: cubic-bezier(.35, 0, .25, 1);
```
simeonoff, line 143: Is this the same as the angular implementation?

src/components/tabs/themes/tabs.base.scss
```diff
@@ -1,116 +1,155 @@
...
 
 :host {
-    display: flex;
-    flex-direction: column;
+    // Scroll buttons
+    --scroll-btn-size: #{rem(48px)};
+    --tabs-width: 100%;
+    --tab-max-width: #{rem(360px)};
+    --tab-min-width: #{rem(90px)};
+    --_border-size: #{rem(1px)};
```
simeonoff, line 12: It may be prudent to include these properties as part of the component theme schema as they need to be in sync with Ignite UI for Angular. Also, I see that some of the properties are used only once, I don't see why they need to be defined is CSS variables. I prefer the value to stay close to the rule in that case.

src/components/tabs/themes/tabs.base.scss
```diff
@@ -1,116 +1,155 @@
...
     }
 }
 
-[part='headers'] {
-    display: flex;
-    overflow: hidden;
-    min-height: rem(42px);
-    flex: none;
+[part~='scrollable'] {
+    /* stylelint-disable */
```
simeonoff, line 58: What stylelint rule are you disabling here? When possible disable only specific rules and only if specifically needed.

src/components/tabs/themes/tab.base.scss
```diff
@@ -53,3 +65,30 @@
...
 }
+
+[part~='tab-body'] {
+    display: none;
+    position: sticky;
+    inset-inline-start: 0;
+    width: var(--tabs-width);
+    grid-row: 3;
+    grid-column: 1 / -1;
+    max-width: var(--_ig-tabs-width);
```
simeonoff, line 76: I cannot see where this CSS variable is declared and the naming is somewhat confusing considering it's so close to the one a few lines above named --tabs-width.

src/components/tabs/themes/shared/tab/tab.fluent.scss
```diff
@@ -3,14 +3,28 @@
 
 $theme: $fluent;
 
-:host {
-    ::slotted(igc-icon) {
-        color: var-get($theme, 'item-icon-color');
-    }
+[part='tab-header'] {
+    transition: all .3s cubic-bezier(.35, 0, .25, 1);
```
simeonoff, line 7: I see the timing function declared in several places. Questioning if this is synced with the implementation in Ignite UI for Angular, and whether we can use one of the included timing functions from the theming package. Else, I would prefer that this value is synced between the two libraries and declared only once, in the common file for instance.

src/components/tabs/themes/tabs.base.scss
desig9stein: Yes, it is used in the exact same way. I know it's not optimal, but I decided not to deal with it in this PR.

src/components/tabs/themes/tab.base.scss
desig9stein: The value for --_ig-tabs-width is dynamically calculated in TypeScript and represents the width of the ig-tabs component. It is required to ensure the position: sticky behavior for the body works properly. Let me know if I need to change the name, for now, I will add a comment to explain its purpose.

src/components/tabs/themes/shared/tab/tab.fluent.scss
desig9stein: We don't have that exact bezier function. The closest one is $in-sine: cubic-bezier(0.47, 0, 0.745, 0.715). Let me know if you'd like me to update it. I’d prefer to handle it in a separate PR after this one is merged, as it will also require updating Angular.

src/components/tabs/themes/tabs.base.scss
```diff
@@ -101,12 +96,12 @@
 
     [part~='scrollable'] {
         /* stylelint-disable */
```
simeonoff, line 98: I guess you want the same stylelint rule as the one a few lines above?

ChronosSF:
This requires patching the breaking changes so that users can still use the component the old way but get deprecation warnings for one release.

simeonoff:
@desig9stein Could you please take a look at the implementation once again with the latest theming package and see if there's anything outstanding? There may be some outdated designs bug-fixes that will need to be retrofitted in this PR so that we don't introduce regressions.

desig9stein (approved):

rkaraivanov (approved):

PR #1354, opened 2024-08-19
chore(*): Do not use spread operator for object in icon service.

MayaKirova:
Issue with that is that if the object is more complex , like the classes it gets transformed through in react: 

```
export class IconMeta extends Base {
	static $t: Type = markType(IconMeta, 'IconMeta');
	private _name: string = null;
	get name(): string {
		return this._name;
	}
	set name(value: string) {
		this._name = value;
	}
	private _collection: string = null;
	get collection(): string {
		return this._collection;
	}
	set collection(value: string) {
		this._collection = value;
	}
}
```

Then the name and collection props are not populated in the spread object. Hence they get lost.


damyanpetev:
@MayaKirova Ah that was the issue with it? Mildly infuriating effect of the wrappers 😑

Question: Instead of adjusting the source, could this be avoided by marking the IconMeta to avoid the wrap? Something like the `tsPlainInterface` tag?

MayaKirova:
> @MayaKirova Ah that was the issue with it? Mildly infuriating effect of the wrappers 😑
> 
> Question: Instead of adjusting the source, could this be avoided by marking the IconMeta avoid the wrap? Something like the `tsPlainInterface` tag?

@damyanpetev I did mark it as an interface [here](https://infragistics.visualstudio.com/NetAdvantage/_git/dev-tools/pullrequest/13612) since it makes more sense for it to be one anyway, but it still goes through some transformations making it into a IconMeta instance before passing to the wc component. So it didn't fix it.

damyanpetev:
> I did mark it as an interface here since it makes more sense for it to be one anyway, but it still goes through some transformations making it into a IconMeta instance before passing to the wc component. So it didn't fix it.

Well that's unfortunate, was worth a shot :)

I realize this might be minor and future plans for the wrappers might obsolete the need for this fix, but _technically_ the class does satisfy the `IconMeta` interface, except we can't spread it of course. Most usage is still likely going to be objects, but since we have no control over incoming params on public API, should we in general avoid assuming the type impl?
In that case this can be a proper fix covered with a test so the change is secured. // CC @rkaraivanov

dkamburov (approved):

damyanpetev (approved):

PR #1357, opened 2024-08-21
fix(radio): Initial checked state through DOM attribute

### LINKED ISSUE #1356: [Bug]: Initially checked radio by attribute throws error when not being last sibling.

Radio & Radio Group

### Description

The following error is thrown when there is initially checked radio using the `checked` attribute and it is not the last radio sibling. This is because it cycles through the sibling elements and the ones with higher index than the checked radio are not yet initialized to call the `updateValidity` method:

TypeError: radio.updateValidity is not a function
    at IgcRadioComponent._updateUncheckedState (radio.js:137:1)
    at set checked (radio.js:58:1)
    at IgcRadioComponent.set [as checked] (reactive-element.js:307:1)
    at HTMLElement._updateCheckedState (radio.js:124:1)
    at set checked (radio.js:58:1)
    at HTMLElement.set [as checked] (reactive-element.js:307:1)
    at HTMLElement._$attributeToProperty (reactive-element.js:662:1)
    at HTMLElement.attributeChangedCallback (reactive-element.js:606:1)
    at RadioAlignment.connectedCallback (index.ts:45:23)
    at Router.displaySample (router.ts:60:34)

### Reproduction

1. Run the sample [radio.zip](https://github.com/user-attachments/files/16679544/radio.zip)
2. Observe the console log

### Workaround

Set the checked attribute manually through code.

### Is this a regression?

### Affected versions

5.0.0-rc.1 - maybe earlier

Any

rkaraivanov:
* Radios in a named group will now wait for the first update before resovling the group state (checked, invalid, etc).
* Binding a value through an igc-radio-group will now set the default value for the group restoring it on form reset.

Closes #1356


skrustev:
Am I missing something or there's another issue now, because it throws pretty much the same error, but the call stack is a bit different. I am using the same sample from the issue with build from this branch:

TypeError: radio.updateValidity is not a function
    at IgcRadioComponent._updateUncheckedState (radio.ts:250:13)
    at set checked (radio.ts:135:55)
    at IgcRadioComponent.set [as checked] (reactive-element.js:307:1)
    at HTMLElement._updateCheckedState (radio.ts:233:20)
    at set checked (radio.ts:135:26)
    at HTMLElement.set [as checked] (reactive-element.js:307:1)
    at HTMLElement._$attributeToProperty (reactive-element.js:662:1)
    at HTMLElement.attributeChangedCallback (reactive-element.js:606:1)
    at registerComponent (register.ts:15:20)
    at IgcRadioComponent.register (radio.ts:59:22)

rkaraivanov:
> For some reason the issue is not present in the storybook. Should the webcomponents be registered some different way?

No, there are no changes made to the registering mechanism of the components in the package.
Can you test with the latest state of this PR? The setter should "wait" until the radio is "upgraded" to a web component before executing any logic which is dependent on being a web component.

I'm not sure about the storybook as well. It fooled me as well, I guess there is a bit of asynchronous/batching logic in it when rendering the stories.

skrustev:
Looks good, just something that I noticed while checking out the Radio Group. Since this is a separate component I kinda expected the radios to work with a simple:

```
<igc-radio-group>
  <igc-radio>Apple</igc-radio>
  <igc-radio>Banana</igc-radio>
  <igc-radio>Mango</igc-radio>
  <igc-radio>Orange</igc-radio>
</igc-radio-group>
```

Although it appears that the user is also required to set `name` for all the radios. I know that's how it is done natively for radios but wasn't expecting it when there's a specific Radio Group component. Just an observation. Maybe just mentioning it in the docs would be also ok.

rkaraivanov:
> Although it appears that the user is also required to set `name` for all the radios. I know that's how it is done natively for radios but wasn't expecting it when there's a specific Radio Group component. Just an observation. Maybe just mentioning it in the docs would be also ok.

Either that or to set a name property on the group itself which will propagate for the radios. Anyway, you can file an issue for the documentation topic in the relevant repository.

src/components/radio-group/radio-group.spec.ts
```diff
@@ -194,6 +195,138 @@ describe('Radio Group Component', () => {
...
+                <igc-radio value="orange" checked>Orange</igc-radio>
+              </igc-radio-group>
+            </form>
+          `);
+          group = form.querySelector(IgcRadioGroupComponent.tagName)!;
+          radios = Array.from(form.querySelectorAll(IgcRadioComponent.tagName));
+          setFormListener();
+
+          // The first checked member of the group takes over
+          expect(first(radios).checked).to.be.true;
```
damyanpetev, line 266: Don't we usually do the last? As in the last command to request selection overrides the previous states

damyanpetev (approved):

PR #1360, opened 2024-08-21
fix(icon-registry): broadcast sync external icon refs only

### LINKED ISSUE #1359: [Bug]: If wc internal set icon matches with grid icon, grid icon is lost.

Icon(s)

### Description

Some of the icons from the internal collection match by name with the ones in the angular element's grid. So if you add any icon to the page, some of the grid icons will not load correctly.

### Reproduction

1. Run the attached sample: 
[sample.zip](https://github.com/user-attachments/files/16692269/sample.zip)

Notice that some of the icons in the pager are missing: 
image

### Workaround

N/A

### Is this a regression?

### Affected versions

5.0.0-rc.1

N/A

MayaKirova:
Closes #1359


src/components/icon/icon-state.broadcast.ts
```diff
@@ -34,10 +34,22 @@ export class IconsStateBroadcast {
     this.send({
       actionType: ActionType.SyncState,
       collections: this.getUserSetCollection(this.collections).toMap(),
-      references: this.refsCollection.toMap(),
+      references: this.getRefsSetCollection(this.refsCollection).toMap(),
     });
   }
 
+  private getRefsSetCollection(collections: IconsCollection<IconMeta>) {
```
damyanpetev, line 41: For consistency this can also be `getUserRefsCollection` kinda like `getUserSetCollection` just to be clearer what we're filtering for.

damyanpetev (commented):
Also would really like this fix to be backed by a test or two - either `setRefsByTheme` or calling `setIconRef` without external (assuming there are already tests with) should produce no broadcast events and the refs should not be present in the synced state

damyanpetev (approved):

PR #1363, opened 2024-08-22
input-group - update label typography

desig9stein:
Closes: [#14649](https://github.com/IgniteUI/igniteui-angular/issues/14649)


simeonoff (commented):
@desig9stein After comparing against Angular the label is missing bottom margin.

simeonoff (approved):

PR #1389, opened 2024-09-17
refactor(form-mixin): Refactor form components for native behavior alignment

### LINKED ISSUE #1386: Input bound to local variable does not reset correctly

Input

### Description

Resetting an input in a form does not reset to initial value. Initial value of the input is bound to a variable. The variable is initialized via call to an end point.

### Reproduction

1. Add variable and initialize it via call to an end point.
2. Add form.
3. Inside the form add an `igc-input` and bound its value to the variable.
4. Add `reset` type button in the form.
5. Run application.
6. Change the input value.
7. Click the reset button.

Code should look like this:
```html
  constructor() {
    super();
    northwindABService.getCustomerDto('ALFKI').then(data => this.customer = data);
  }

  @state()
  private customer?: CustomerDto;

  render() {
    return html`
      <form @submit="${this.submit}" class="column-layout form">
        <igc-input value="${ifDefined(this.customer?.customerId)}" name="customerId"></igc-input>
        <igc-button variant="outlined" type="reset" class="button">Reset<igc-ripple></igc-ripple>
      </form>
    `;
  }
```

### Workaround

Not workaround but if I use `input` reset works correctly. Also, if I bound the `igc-input` to a constant reset works correctly.

### Is this a regression?

### Affected versions

4.11.1

All

rkaraivanov:
- Aligned form components to mimic native counterparts when setting `value`/`checked` states via `setAttribute()`, ensuring internal `defaultValue` is updated accordingly.
- Removed obsolete code.
- Consolidated code into a more efficient API.
- Renamed internal methods for clarity.

Closes #1386


rkaraivanov:
@wnvko 
Feel free to test more form components if you have the time 😁

wnvko:
@radomirchev 
Rating with no initial value does not reset after user selects some value. Having rating in form like this:
```html
<form action="">
  <igc-rating name="my-rating"></igc-rating>
  <igc-button type="reset">Reset</igc-button>
</form>
```
If I choose some value and then click reset button nothing happens - rating retains its value. This bug does not happen if rating has initial value, or if rating is `single`.

rkaraivanov:
> @radomirchev Rating with no initial value does not reset after user selects some value. Having rating in form like this:
> 
> ```
> <form action="">
>   <igc-rating name="my-rating"></igc-rating>
>   <igc-button type="reset">Reset</igc-button>
> </form>
> ```
> 
> If I choose some value and then click reset button nothing happens - rating retains its value. This bug does not happen if rating has initial value, or if rating is `single`.

I suppose that replay was for me.
Thanks for catching the issue, it should be fixed now.

src/components/common/mixins/forms/types.ts
```diff
@@ -0,0 +1,160 @@
+import type { Validator } from '../../validators.js';
+
+export type FormRestoreMode = 'autocomplete' | 'restore';
+export type FormValueType = string | File | FormData | null;
+
+declare class BaseFormAssociatedElement {
+  public static readonly formAssociated: boolean;
+
+  private __internals: ElementInternals;
```
damyanpetev, line 9: PS: This is probably redundant since it's private and can't reflect on classes using the mixin in any way.
Still a bit iffy on the declared classes since they can't enforce the implementation, but I get why we need them.

damyanpetev (approved):
Other than the minor comment on the internal form mixin impl, LGTM

PR #1391, opened 2024-09-18
Chip - update indigo theme

desig9stein:
Related to https://github.com/IgniteUI/igniteui-theming/issues/326

### DIFF
--- a/src/components/chip/themes/dark/chip.indigo.scss
+++ b/src/components/chip/themes/dark/chip.indigo.scss
@@ -7,3 +7,8 @@ $theme: $indigo;
 :host {
     @include css-vars-from-theme(diff(light.$base, $theme), 'ig-chip');
 }
+
+:host([selected]) button[disabled],
+:host([variant='primary']) button[disabled] {
+    color: contrast-color($color: 'primary', $variant: 900, $opacity: .2);
+}
--- a/src/components/chip/themes/light/chip.indigo.scss
+++ b/src/components/chip/themes/light/chip.indigo.scss
@@ -6,3 +6,8 @@ $theme: $indigo;
 :host {
     @include css-vars-from-theme(diff($base, $theme), 'ig-chip');
 }
+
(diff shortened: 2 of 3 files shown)


andiesm813:
Everything looks good here too. Thanks @desig9stein

yradoeva:
There are some things I noticed during validation:

**Light & Dark mode**
1. L, M and S size - the width of the chip is with 2 px larger because the border is implemented outside vs. inside in the Design kit.
In that case left and right paddings should be:
L - 7px
M - 5px
S - 3px
(screenshot is from M)
image

**Light mode**
2. **Hover** - border color should be grays.500 (not grays.400)
image

3. **Selected - disabled:** text color should be white 40% (not grays.900 20%)
image

4. **Primary - disabled:** text color should be white 40% (not white)
image

5. **Primary + Selected - disabled:** text color should be white 40% (not white)
image


**Dark mode**
6. **Hover** - border color should be grays.200 (not grays.100)
image

7. **Primary - disabled:** text color should be white 20% (not white)
image

8. **Primary + Selected - disabled:** text color should be white 20% (not white)
image

andiesm813:
@desig9stein i dont know if you fixed it yet, but i dont see the issue nº 2... 
image

Some of the opacity issues might be related to the addition of opacity on the disabled component, yes @desig9stein ? So the text color being white for some of the variant chips (info, success, warning and error) is correct according to the handoff.  But the primary chip didnt follow these rules, because it looked identical to the selected chip... so i kept the same criteria for the primary chip. But this may be confusing.... 

The other variants are fine on disabled.... maybe i still need to fix the primary chip, to match the same behavior as the other colored variants?

image

yradoeva:
There is only thing left to be fixed in **Dark mode**:

**Default Selected - disabled
Primary - disabled
Primary Selected - disabled**
text and icon color should be white 20% (not white 40%)
image

simeonoff (approved):

PR #1399, opened 2024-09-25
refactor(badge): update font-size and padding

simeonoff:


SisIvanova (changes requested):
There is an issue with the bootstrap square badge
Screenshot 2024-09-25 at 2 24 55 PM

SisIvanova (dismissed):

PR #1402, opened 2024-09-25
[WIP] Tile manager component implementation

### LINKED ISSUE #1379: Tile Manager Component

- [x] Research
- [x] Specification - https://github.com/IgniteUI/igniteui-webcomponents/wiki/Tile-Manager-Specification
- [x] API Documentation
- [x] Design implementation

mtsvyatkova:
Closes #1379


src/components/tile-manager/tile-manager.ts
```diff
@@ -173,6 +180,11 @@ export default class IgcTileManagerComponent extends EventEmitterMixin<
     });
   }
 
+  public override connectedCallback() {
+    super.connectedCallback();
+    this.updateIsAnyTileMaximized();
+  }
+
```
rkaraivanov, line 187: Not really needed

rkaraivanov (changes requested):

src/components/tile-manager/tile.ts
```diff
@@ -291,42 +292,57 @@ export default class IgcTileComponent extends EventEmitterMixin<
...
+    e.dataTransfer!.setDragImage(this._dragImage, offsetX, offsetY);
     e.dataTransfer!.effectAllowed = 'move';
+  }
+
+  private async handleDragStart(e: DragEvent) {
+    await this.assignDragImage(e);
 
     this.emitEvent('tileDragStart', { detail: this });
-    this._isDragging = true;
+    this._dragGhost = this.ghostFactory();
```
rkaraivanov, line 320: What's the idea behind this? We need an additional _ghost_ element for?

src/components/tile-manager/tile.ts
```diff
@@ -291,42 +292,57 @@ export default class IgcTileComponent extends EventEmitterMixin<
...
 
-    e.dataTransfer!.setDragImage(this._tileContent, offsetX, offsetY);
+    this._dragImage = this.cloneNode(true) as HTMLElement;
+    Object.assign(this._dragImage.style, {
+      width: compStyles.width,
+      height: compStyles.height,
+    });
+    this._dragImage.style.position = 'absolute';
+    this._dragImage.style.top = '-99999px';
+    this._dragImage.style.left = '-99999px';
```
rkaraivanov, line 308: You can move those assignments in the object assign call above

src/components/tile-manager/tile.ts
```diff
@@ -291,42 +292,57 @@ export default class IgcTileComponent extends EventEmitterMixin<
     });
   }
 
-  private handleDragStart(e: DragEvent) {
+  private async assignDragImage(e: DragEvent) {
```
rkaraivanov, line 295: I don't see any `await` here, this can be a synchronous method.

rkaraivanov (commented):
@mtsvyatkova 
Can you address some of the comments around the drag and drop functionality?

src/components/tile-manager/tile.ts
mtsvyatkova: The additional ghost is shown on the tiles' place, showing a visual clue of the current tile's position (or, in swap mode, the position where the drag started). I believe it came up as a design requirement and a solution to add some transition to the dragging operation.

gedinakova:
@rkaraivanov @onlyexeption The following error is logged in the console on attempt to resize a maximized tile (even though it's not showing the resize handler). 
image

gedinakova:
@rkaraivanov The resize handle is now visible when a tile is in a fullscreen state. If it gets clicked the tile gets blurred although the buttons are still interactable. Once restored to normal state if you click on the maximize button the layout is broken.

[screen-capture (2).webm](https://github.com/user-attachments/assets/f6a38c2a-5795-4cd3-bab9-8d00bfda44d5)

gedinakova:
Resize indicators bug

1. Go to "Dynamic Tiles" story
2. Enable resizeMode on hover
3. Hover over the first tile
4. Mouse down and drag towards the second tile
5. Note the resize indicators over the second tile

Result: The indicators get selected and look very odd. Hover over another tile and then back to the second one. The indicators are still like that.

image

gedinakova:
[Bug] Maximize & fullscreen actions are broken after resizing a tile

Steps to reproduce:

1. Load Fin Dashboard story
2. Resize a tile
3. Attempt to maximize/fullscreen it

Result: The tile is moved to the top left, its state is changed but its size remain the same.

image

Expected result: The tile should be successfully maximized/fullscreened

gedinakova:
[Bug] All tiles on the row flash quickly when resizing a tile

Steps to reproduce:
1. Go to Serialization story & enable resizing
2. Resize Tile2 taller
3. Resize Tile3 taller

Result:
Tile1 is also quickly resized to the same size as Tile3 then restored back to its own size.

Expected result:
Other tiles should not be affected.

gedinakova:
[Bug] Tiles overlap after resizing higher than the initial TM height

1. Go to Dynamic tiles story & enable resizing (Make sure the TM is wide enough so it has only one row of tiles)
2. Resize the first tile higher
3. Resize the second tile wider

Result: The second tile overlaps the first.

[screen-capture (14).webm](https://github.com/user-attachments/assets/2caee07c-42c6-46e5-a183-d2909a49cd6a)

Expected result: Tiles should be positioned without overlapping.

gedinakova:
[Bug] Tiles height is not properly restored on loadLayout

1. Go to Serialization story
2. Save the layout
3. Resize one of the tiles higher
4. Load the saved layout

Result: The tiles remain with the greater height.

Expected result: The tiles' height should also be restored.

gedinakova:
[Bug] Hide maximize button when a tile is in fullscreen mode (as discussed with Dani in the last meeting).

gedinakova:
[Bug] All tiles' height on the row is affected by one tile resizing
1. Load the default story
2. Set column count to 6 so the tiles are on 2 rows
3. Resize a tile from the second row higher
4. Resize a tile from the first row higher

Result: All tiles on the first row get resized.

[screen-capture (15).webm](https://github.com/user-attachments/assets/e3c6113f-342a-4055-8068-5f5ac7d2fb05)

Expected result: The resize operation should affect a single tile.

yradoeva:
[Bug] The tile doesn't get the correct visual style when being resized after being drag&dropped

1. Load the Fin Dashboard 1 story
2. Enable resizing and dragging
3. Drag & drop the first tile to change its position
4. Resize the same tile
5. Resize another tile, which is not dragged

Result: the drag&dropped tile is not getting the correct style (idle state with 30% opacity) but stays as in idle. Tiles which have not been drag&dropped are looking ok when being resized.

Expected: the tile being resized becomes transparent with 30% opacity (as per design). 

https://github.com/user-attachments/assets/c2249bb0-f9c6-4a2a-b30c-76f3628f806e

yradoeva:
[Bug] A tile with disabled resize should not have a border, styled as for resizing, while in resizeMode:always. It should stay in idle state.
Screenshot 2025-03-06 at 11 31 12

yradoeva:
[Bug] The tile keeps its hover state after maximizing/minimizing even without being hovered while in resizeMode:hover & resizeMode:always.

1. Load the Default story
2. Set the resizeMode to hover
3. Maximize Tile 2, move the mouse out of the tile manager - observe the tile border and the maximize button
4. Minimize Tile 2 - observe the tile border and the maximize button
5. Set the resizeMode to always
6. Maximize Tile 2, move the mouse out of the tile manager - observe the tile border and the maximize button
7. Minimize Tile 2 - observe the maximize button

Result:
- after 3 - the tile stays with border as for resizing, the maximize button stays hovered
Screenshot 2025-03-06 at 11 53 06

- after 4 - the tile stays with border as for resizing, the maximize button stays hovered
Screenshot 2025-03-06 at 11 53 28

- after 6 - the tile stays with border as for resizing, the maximize button stays hovered
- after 7 - the maximize button stays hovered


Expected:
The tile and the maximize button should be in idle state after being maximized/minimized. Only on hover they should be in hover state. When maximized, the tile should not be resizable and therefore should not have a border styled as for resizing.

https://github.com/user-attachments/assets/59234ade-2256-4a4b-90ca-8926427486c9

gedinakova:
@rkaraivanov @onlyexeption 
This is definitely not a showstopper, but if possible we can revisit it after the release.
Currently, horizontal resizing seems a little bit too restrictive for tiles positioned after the first column. Meaning, we allow tiles to be resized so that they span through all columns of the current grid layout. However, this is only valid for tiles starting in the first column. All tiles starting from subsequent columns would be allowed to span n - columnIndex columns. Given the fact that if reordered the same tiles would be allowed to span more columns, isn't it better if we reconsider the current implementation and allow all tiles to increase their span to the current maximum?

gedinakova:
[Bug] Resizing bottom row tiles higher happens after a very small treshold

1. Load Dynamic tiles story
2. Resize the browser so the tiles are on 2 rows
3. Drag the bottom (corner) resize handle down so the ghost's bottom is 1-2 pixels below the tile manager's bottom
4. Release the mouse

Result: The tile is resized to occupy +1 rows.

[screen-capture (17).webm](https://github.com/user-attachments/assets/5732ee0c-59c4-4f8f-bf71-a50b29388772)

Expected result: The tiles on the bottom row should be resized only after the resize ghost has been resized over the treshold as tiles on the upper rows do.

SisIvanova:
[Bug] When styling the resize indicators using the corresponding CSS parts, styles are being applied to both custom and default indicators if they are used together. Instead, we should have a way to distinguish them and change the design of only custom or default indicators. 
Screenshot 2025-03-10 at 1 35 13 PM

gedinakova:
[Bug] Drag ghost is not hidden if both mouse buttons are clicked during dragging

1. Load the Dynamic tiles story
2. Enable dragging
3. Start dragging a tile
4. Click both mouse buttons while keeping dragging (it may take a few tries)

Result: The ghost freezes at some position and remains there after releasing the mouse button.

[screen-capture (18).webm](https://github.com/user-attachments/assets/981e3788-81d9-4c70-a019-8afb2cbad95e)

Expected result: Right-click should not affect the ghost.

rkaraivanov (approved):

PR #1408, opened 2024-09-26
Themes: Indigo updates for the Chip and Tabs

simeonoff:


yradoeva:
**Tabs**
**Light and Dark mode**

1. Bottom border of the active tab should be 3px high (not 2px)
Background
<br/>
<br/>
2. Removable tabs are implemented with contained icon buttons - they should be flat, but as this is valid for all themes, I suppose it will be logged as an issue and fixed at a later stage. I just wanted to note it here.
Pasted Graphic 80

SisIvanova (approved):

PR #1421, opened 2024-10-07
tests: Refactor validation container integration scenarios

rkaraivanov:


src/components/checkbox/checkbox.spec.ts
```diff
@@ -389,39 +389,15 @@ describe('Checkbox', () => {
...
-      await checkValidationSlots(element, 'invalid');
+    it('', async () => {
+      const testParameters: ValidationContainerTestsParams<IgcCheckboxComponent>[] =
+        [
+          { slots: ['valueMissing'], props: { required: true } }, // value-missing slot
+          { slots: ['customError'] }, // custom-error slot
+          { slots: ['invalid'], props: { required: true } }, // invalid slot
+        ];
+
+      runValidationContainerTests(new IgcCheckboxComponent(), testParameters);
```
damyanpetev, line 400: Eh, if the only goal is to pass the tag you could do that, but to enforce the impl you could also:
```suggestion
      runValidationContainerTests(IgcCheckboxComponent, testParameters);
```
With the `runValidationContainerTests` param as  `element: Constructor<T> & IgniteComponent` to get the same behavior.

Then again I was also okay with the old fixture approach, just reducing the repetition a bit. Any benefits of `fixture` over basic create element?

damyanpetev (approved):

PR #1439, opened 2024-10-17
fix(slider): aria-label issue

### LINKED ISSUE #1410: [Bug]: Incorrect aria-label implementation on igc-slider component

Slider (single/range)

### Description

There is an issue (enhancement) with the igc-slider component where the aria-label attribute is applied to both the igc-slider and its inner thumb element. This is not correct because the aria-label should be applied only to the thumb, which is the focusable part of the slider. When navigating via keyboard (tabbing), users focus directly on the thumb and not the slider container itself.

### Reproduction

To observe the issue:

    1. Include an igc-slider component in your project.
    2. Inspect the rendered HTML in the browser.
    3. Notice that both the igc-slider and its inner thumb element have the aria-label attribute applied.

### Workaround

### Is this a regression?

### Affected versions

5.0.2

Browser: All modern browsers (e.g., Chrome, Firefox, Edge)
Operating System: All platforms (e.g., Windows, macOS, Linux)

SisIvanova:
Closes #1410

### DIFF
--- a/src/components/slider/slider-base.ts
+++ b/src/components/slider/slider-base.ts
@@ -548,7 +548,11 @@ export class IgcSliderBaseComponent extends LitElement {
     return html`<div part="ticks">${this._renderTicks()}</div>`;
   }
 
-  protected renderThumb(value: number, ariaLabel?: string, thumbId?: string) {
+  protected renderThumb(
+    value: number,
+    sliderThumbLabel?: string,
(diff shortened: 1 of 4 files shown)


simeonoff:
@SisIvanova we will have to defer this fix as after some discussion with @rkaraivanov, we would like to implement a different solution that will solve this and other issues in the component.

PR #1459, opened 2024-11-01
fix(tabs): Nested tabs selection behavior

### LINKED ISSUE #713: An igc-tabs inside an igc-tab-panel (nested tab components) doesn't work well

rkaraivanov:
Closes #713


src/components/tabs/tabs.ts
```diff
@@ -121,17 +121,23 @@ export default class IgcTabsComponent extends EventEmitterMixin<
   private _mutationCallback({
     changes: { attributes, added, removed },
   }: MutationControllerParams<IgcTabComponent>) {
-    this.setSelectedTab(attributes.find((tab) => tab.selected));
+    const ownAttributes = attributes.filter(
+      (tab) => tab.closest(this.tagName) === this
+    );
+    const ownAdded = added.filter(({ target }) => target === this);
+    const ownRemoved = removed.filter(({ target }) => target === this);
```
damyanpetev, line 128: Shouldn't we check these with `target.closest(this.tagName) === this` still? Just like the attributes. In most cases these should be direct `igc-tab` children, but technically those can be wrapped in containers, which will shift the target.

src/components/carousel/carousel.ts
```diff
@@ -160,8 +160,9 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
     if (activeSlides.length <= 1) {
       return;
     }
-
-    const idx = this.slides.indexOf(last(added.length ? added : attributes));
+    const idx = this.slides.indexOf(
+      added.length > 0 ? last(added).node : last(attributes)
```
damyanpetev, line 164: Just a minor nit, but this being the same check as in the button group, can those be the same - i.e. `added.length ? ...`

damyanpetev (approved):

PR #1470, opened 2024-11-11
refactor: Added support for string-based property values for date based components

### LINKED ISSUE #1467: [Bug]: DatePicker throws error when the data is string

Date picker

### Description

When the Date comes from  an endpoint in string format , DatePicker throws error.

### Reproduction

In a forms scenario where the form is bound to an endpoint with Date fields, attempting to run the application result in the following error. 
image

### Workaround

no

### Is this a regression?

### Affected versions

latest

Chrome

rkaraivanov:
This commit adds support for passing in string based date values to date-bound properties of calendar, date-time input and date picker components. The component will now try to parse the passed in value without directly throwing an exception.
Unified the type declarations between date-bound component.

Closes #1467


damyanpetev (approved):
Should probably plan to revise the date utils down the line, now that some of the parsing is shared with the calendar. Other than that, LGTM :)

damyanpetev:
@Marina-L-Stoyanova Just to check - have you tested this change in some more complete scenario than the repo sample for the issue? As in some of the AB form scenarios - asking specifically because we discussed the code produced for nullable variables or request-bound would produce optional values, see https://github.com/IgniteUI/igniteui-webcomponents/pull/1470#discussion_r1838319108 and with the latest version those would result in type check errors:
image
Those are visible with a VS Code Lit plugin or lit eslint plugin and strict checks and this PR should clear those up, but I wanted to verify that.

Marina-L-Stoyanova:
> @Marina-L-Stoyanova Just to check - have you tested this change in some more complete scenario than the repo sample for the issue? As in some of the AB form scenarios - asking specifically because we discussed the code produced for nullable variables or request-bound would produce optional values, see [#1470 (comment)](https://github.com/IgniteUI/igniteui-webcomponents/pull/1470#discussion_r1838319108) and with the latest version those would result in type check errors: image Those are visible with a VS Code Lit plugin or lit eslint plugin and strict checks and this PR should clear those up, but I wanted to verify that.


Yes, everything looks ok. I have tested it with form application that has date picker inputs. I installed the suggested plugin and it is working as expected.

PR #1486, opened 2024-11-18
fix(chip): styling discrepancies

### LINKED ISSUE #1458: Chip prefix and suffix expand the chip when empty

Chip

### Description

When the prefix and suffix are not used, they're present in the dom and empty, but expand the chip. When prefix or suffix slot is empty, it should not expand the chip, the chip should be left with just it's paddings.

Screenshot 2024-11-01 at 8 31 18

### Reproduction

Check out the chip story, explore different themes and sizes and compare with UI kits.

### Workaround

I have not found a workaround

### Is this a regression?

### Affected versions

all

all

SisIvanova:
Closes #1458


didimmova:
There are still some misalignments colorwise in the foregrounds in different variants. Please check this out with the design team if possible.

src/components/chip/themes/shared/chip.common.scss
```diff
@@ -45,4 +41,9 @@ $theme: $material;
     ::slotted(igc-circular-progress) {
         --diameter: #{sizable(rem(14px), rem(18px), rem(24px))};
     }
+
+    igc-icon,
+    ::slotted(igc-icon) {
+        --size: #{sizable(rem(18px), rem(18px), rem(24px))};
+    }
```
simeonoff, line 48: As far as I remember the size of the icon in the chip should not be modified by the chip component. Rather, it should follow the default icon sizes.

simeonoff (commented):
I believe we shouldn't be modifying the icon sizes in the context of the chip component.

src/components/chip/themes/shared/chip.common.scss
SisIvanova: Correct, the issue is in the UI Kit.

simeonoff (approved):

PR #1492, opened 2024-11-22
refactor(progressbar): Animate the value change using CSS counters and custom CSS props

### LINKED ISSUE #1441: [Change Request]: Animate the value change using CSS counters and custom CSS props

Progress bar (circular/linear)

### Description

Let's refactor the animation of the value (now done using JavaScript with requestAnimationFrame) to a pure CSS solution by utilizing a [custom CSS property of type integer](https://developer.mozilla.org/en-US/docs/Web/CSS/@property) in combination with animating a [CSS counter](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters).

desig9stein:
closes #1441


rkaraivanov (dismissed):

src/components/progress/themes/circular/shared/circular.progress.common.scss
```diff
@@ -26,27 +26,22 @@ $theme: $material;
     stroke: var-get($theme, 'base-circle-color');
 }
 
+// Mixin for Gradient Colors
+@mixin gradient-variant($variant, $color) {
+    :host([variant='#{$variant}']) [part~='gradient_start'],
+    :host([variant='#{$variant}']) [part~='gradient_end'] {
+        stop-color: $color;
+    }
+}
```
simeonoff, line 35: I like that you're optimizing this! Can we instead use a CSS variable to set the step-color for the different variants to reduce the amount of generated CSS?

src/components/progress/themes/animations.scss
```diff
@@ -1,37 +1,41 @@
...
 
+// Primary animation
 @keyframes indeterminate-primary {
     0% {
         transform: translateX(0);
     }
 
     20% {
-        animation-timing-function: $indeterminate-primary-translate-step-2;
+        animation-timing-function: map.get($easing-curves, primary-translate-start);
```
simeonoff, line 33: Wrap this in a function that uses the map.get call internally referencing the $easing-curves. Say the function is named `timing-function`, then your call will simply be:

```suggestion
        animation-timing-function: timing-function(primary-translate-start);
```

src/components/progress/themes/circular/circular.progress.base.scss
```diff
@@ -61,30 +79,39 @@
...
 
-    --percentage: 0;
-    --duration: 500ms;
+    [part~='fill'] {
+        stroke-dashoffset: calc(#{$circumference} * 2);
+        animation: indeterminate-accordion 1.5s cubic-bezier(0, 0.085, 0.68, 0.53) $animation-direction infinite;
+    }
 }
 
+:host-context([dir='rtl']),
```
simeonoff, line 100: As far as I am aware, this does not yet work in Safari and Firefox?

src/components/progress/themes/circular/shared/circular.progress.bootstrap.scss
```diff
@@ -13,13 +13,13 @@ $theme: $bootstrap;
     font-weight: 700;
 }
 
-[part~='svg'][part~='indeterminate'] {
+[part='svg indeterminate']  {
```
simeonoff, line 16: Are you sure that these are the only parts that can be applied to this element? As soon as any other part is applied the styles will stop working.

src/components/progress/themes/circular/shared/circular.progress.fluent.scss
```diff
@@ -18,3 +18,19 @@ $theme: $fluent;
...
+[part='svg indeterminate'] {
+    animation: none;
+
+    [part~='fill'] {
+        stroke-linecap: round;
+        animation: 2s linear 0s infinite normal none running indeterminate-circular-fluent
+    }
+}
+
+:host-context([dir='rtl']),
```
simeonoff, line 31: Again, if this doesn't work in all browsers, we can't rely on it.

src/components/progress/themes/circular/shared/circular.progress.indigo.scss
```diff
@@ -4,10 +4,10 @@
 $theme: $indigo;
 
 :host {
-    --stroke-thickness: 2px;
+    --stroke-thickness: 3px;
```
simeonoff, line 7:
```suggestion
    --stroke-thickness: rem(3px);
```

simeonoff: Also, are we using this variable elsewhere?

src/components/progress/themes/linear/linear.progress.base.scss
```diff
@@ -1,21 +1,25 @@
...
 @use '../vars' as *;
 
 :host {
-    @extend %host !optional;
+    --track-size: #{rem(4px)};
+    --linear-animation-duration: 2000ms;
+}
+
+[part~='base'] {
+    @extend %base !optional;
```
simeonoff, line 13: Why are you marking this as optional? Is there a case in which this placeholder class doesn't exist?

src/components/progress/themes/linear/linear.progress.base.scss
```diff
@@ -1,21 +1,25 @@
...
     position: relative;
     width: 100%;
     flex-direction: column;
     flex: 1 1 100%;
-
-    --track-size: #{rem(4px)};
-    --linear-animation-duration: 2000ms;
 }
 
+:host-context([dir='rtl'])[indeterminate],
```
simeonoff, line 22: See other comments.

src/components/progress/themes/linear/shared/linear.progress.bootstrap.scss
```diff
@@ -10,56 +11,21 @@ $stripe-size: rem(5px);
...
-    @include stripe-colors(var-get($theme, 'fill-color-warning'), var-get($theme, 'stripes-color'), $stripe-size, 45deg);
-}
-
-[part~='striped'][part~='warning'] {
-    &::after {
-        @include stripe-colors(var-get($theme, 'fill-color-warning'), var-get($theme, 'stripes-color'), $stripe-size, 45deg);
+// Define a mixin for generating striped styles
+@mixin striped-styles($part, $color-key) {
+    [part~='striped'][part~='#{$part}']:not([part~='indeterminate']) {
+        @include stripe-colors(var-get($theme, $color-key), var-get($theme, 'stripes-color'), $stripe-size, 45deg);
```
simeonoff, line 17: Similar to my previous comment, I would appreciate if the background color was made a CSS variable. Also, this ignores the indeterminate state now, effectively changing the behavior. We don't want to have stripes in the indeterminate state.

src/components/progress/themes/linear/shared/linear.progress.common.scss
```diff
@@ -14,102 +15,38 @@ $theme: $material;
...
 
-[part~='striped'][part~='info']:not([part~='indeterminate']) {
-    @include stripe-colors(var-get($theme, 'fill-color-info'), var-get($theme, 'stripes-color'), $stripe-size, -45deg);
-}
-
-[part~='striped'][part~='info'] {
-    &::after {
-        @include stripe-colors(var-get($theme, 'fill-color-info'), var-get($theme, 'stripes-color'), $stripe-size, -45deg);
+// Mixin for striped styles
+@mixin striped-styles($part, $color-key) {
```
simeonoff, line 31: This mixin is duplicated in the bootstrap theme, where I commented that having a CSS variable that changes based on the type (info, success, danger, etc.) will be better).

src/components/progress/themes/shared.scss
```diff
@@ -0,0 +1,22 @@
...
+
+@mixin stripe-colors($primary, $gray, $size, $deg: -45deg) {
+    background: $primary repeating-linear-gradient(
+        $deg,
+        $primary,
+        $primary $size,
+        $gray $size,
+        $gray ($size * 2)
+    );
+}
```
simeonoff, line 22: Having these is not necessary IMO. It Makes you jump between files just to read a few lines. Also, if you make the background property a CSS variable, this mixin is not needed.

src/components/progress/themes/circular/circular.progress.base.scss
rkaraivanov: They don't support it and AFAIK they have no plans to ever do. Not to mention it does not correctly handles `auto` and inherited values.

[`:dir`](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir) is the one which should be used here.

simeonoff:
@desig9stein check the review comments.

simeonoff (approved):

rkaraivanov (dismissed):

src/components/progress/themes/circular/circular.progress.base.scss
simeonoff: @desig9stein This should be fixed.

simeonoff (changes requested):

src/components/progress/themes/circular/shared/circular.progress.indigo.scss
simeonoff: As well as this.

simeonoff (approved):

PR #1518, opened 2024-12-12
feat(textarea): fix material textarea label positioning

### LINKED ISSUE #1506: Textarea - (material) label is not positioned correctly

Textarea

### Description

The label behaves like there is always a value or placeholder
image

### Reproduction

Compare the text area in both libraries

### Workaround

NO

### Is this a regression?

### Affected versions

all

all

didimmova:
Closes #1506


desig9stein:
@didimmova, the default sample is not working correctly

image


if I add this in the story it's working as expected
`placeholder=${args.placeholder || nothing}`

didimmova:
> @didimmova, the default sample is not working correctly
> 
> image
> if I add this in the story it's working as expected `placeholder=${args.placeholder || nothing}`

It's because you've already clicked the "set string" button on the placeholder control and it's standing as an attribute to the tag although it's empty.

desig9stein (dismissed):

desig9stein (approved):

src/components/textarea/themes/shared/textarea.material.scss
```diff
@@ -224,6 +246,18 @@ textarea {
         }
     }
 
+    [part~='container'][part~='filled'],
+    [part~='container'][part~='placeholder'] {
+        [part='notch'] {
+            border-block-start:rem(2px) solid transparent;
```
simeonoff, line 252:
```suggestion
            border-block-start: rem(2px) solid transparent;
```

simeonoff (approved):

PR #1532, opened 2025-01-07
fix(icon-broadcast): Do not sync between wc services, just with angul…

### LINKED ISSUE #1524: [Bug]: CPU usage and Memory consumption are abnormally high when the app is opened in 2, 3, or more tabs

Tree, Other/unknown (please specify in the description)

### Description

CPU usage and Memory consumption are abnormally high when the app with both "Tree" and "Grid" components on its page is opened in 2, 3, or more tabs.

As far as my investigation, it might be caused by an infinite loop consisting of BroadcastChannel message passing. (The channle name is `"ignite-ui-icon-channel"`.)

image

### Reproduction

1. Download and unpack the following zip file.
[📦c-00240606-wc-app1.zip](https://github.com/user-attachments/files/18161789/c-00240606-wc-app1.zip)

2. Run the `npm ci` and `npm run dev` commands.
3. A web browser will open the app with both "Tree" and "Grid" components on its page.
4. Do the "Duplicate tab" on the web browser's tab twice or 3 times.
5. The browser's Memory consumption will increase massively, and CPU usage will remain high. There may be no way to terminate the web browser's process from a Task Manager. 

### Workaround

Although not recommended at all, you can avoid this problem by adding the following script.

```html
<script>
    class BroadcastChannelMock {
        constructor(name) {
            // NOP
        }
        addEventListener(event, callback) {
            // NOP
        }
        postMessage(message) {
            // NOP
        }
    }
    window.BroadcastChannel = BroadcastChannelMock;
</script>
```

However, this workaround may also break other functions and application behavior.

### Is this a regression?

### Affected versions

Failing in 5.x, worked in 4.x

Browser: Chrome 131.0.6778.140 (Official Build) (64bit), Edge 131.0.2903.99 (Official build) (64-bit), Firefox 133.0.3 (64-bit)
OS: Windows 11 Pro 24H2 10.0.26100.2605
Node version: 20.18.0
npm version: 10.8.2

MayaKirova:
…ar-elements.
Closes #1524

### DIFF
--- a/src/components/icon/icon-state.broadcast.ts
+++ b/src/components/icon/icon-state.broadcast.ts
@@ -31,7 +31,11 @@ export class IconsStateBroadcast {
   }
 
   public handleEvent({ data }: MessageEvent<BroadcastIconsChangeMessage>) {
-    if (data.actionType !== ActionType.SyncState) {
+    // no need to sync with other wc icon services, just with angular elements
+    if (
+      data.actionType !== ActionType.SyncState ||
+      data.origin === 'igniteui-webcomponents'
+    ) {
       return;
     }
 
@@ -39,6 +43,7 @@ export class IconsStateBroadcast {
       actionType: ActionType.SyncState,
       collections: this.getUserSetCollection(this.collections).toMap(),
       references: this.getUserRefsCollection(this.refsCollection).toMap(),
+      origin: 'igniteui-webcomponents',
     });
   }
 
--- a/src/components/icon/icon-state.broadcast.ts
+++ b/src/components/icon/icon-state.broadcast.ts
@@ -11,6 +11,7 @@ export class IconsStateBroadcast {
   private channel!: BroadcastChannel | null;
   private collections: IconsCollection<SvgIcon>;
   private refsCollection: IconsCollection<IconMeta>;
+  private static readonly origin = 'igniteui-webcomponents';
 
   constructor(
     collections: IconsCollection<SvgIcon>,
@@ -34,7 +35,7 @@ export class IconsStateBroadcast {
     // no need to sync with other wc icon services, just with angular elements
     if (
       data.actionType !== ActionType.SyncState ||
-      data.origin === 'igniteui-webcomponents'
+      data.origin === IconsStateBroadcast.origin
     ) {
       return;
     }
@@ -43,7 +44,7 @@ export class IconsStateBroadcast {
       actionType: ActionType.SyncState,
       collections: this.getUserSetCollection(this.collections).toMap(),
       references: this.getUserRefsCollection(this.refsCollection).toMap(),
-      origin: 'igniteui-webcomponents',
+      origin: IconsStateBroadcast.origin,
     });
   }
 
--- a/src/components/icon/icon.spec.ts
+++ b/src/components/icon/icon.spec.ts
@@ -9,6 +9,7 @@ import { stub } from 'sinon';
 
 import { defineComponents } from '../common/definitions/defineComponents.js';
 import { first, last } from '../common/util.js';
+import { IconsStateBroadcast } from './icon-state.broadcast.js';
 import IgcIconComponent from './icon.js';
 import {
   getIconRegistry,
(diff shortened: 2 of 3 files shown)


damyanpetev (dismissed):

src/components/icon/icon-state.broadcast.ts
```diff
@@ -31,14 +31,19 @@ export class IconsStateBroadcast {
   }
 
   public handleEvent({ data }: MessageEvent<BroadcastIconsChangeMessage>) {
-    if (data.actionType !== ActionType.SyncState) {
+    // no need to sync with other wc icon services, just with angular elements
+    if (
+      data.actionType !== ActionType.SyncState ||
+      data.origin === 'igniteui-webcomponents'
```
rkaraivanov, line 37: Can we refactor this into a readonly static property on the service instead?

rkaraivanov (changes requested):
LGTM
I was wondering if you can reproduce the bug and cover it in our test suite by creating (several) services that mimic the Angular icon service logic and try to trample over the icon service in the library? If possible, this can be a separate PR.

rkaraivanov (dismissed):

src/components/icon/icon.spec.ts
```diff
@@ -251,6 +253,53 @@ describe('Icon broadcast service', () => {
...
+        }
+      );
+
+      // a peer is requesting a state sync
+      channel.postMessage({ actionType: ActionType.SyncState });
+      await nextFrame();
+
+      // all icon broadcasts must respond once on the channel with their state
+      // so 1 event from the peer + 1 response event with state per broadcast service.
+      expect(evt1.length).to.equal(4);
```
damyanpetev, line 294: Eh, actually it's the 1 from peer + 2 from the other two broadcast services + one from the global service since the previous tests populate the global icon registry (`getIconRegistry()`) which persists and has a broadcast service too :)
That also means this test will fail if ran first or alone with 'expected 3 to equal 4' so you might want to ensure the global service is always initialized as one of the test broadcasts by calling `getIconRegistry()`?

src/components/icon/icon.spec.ts
```diff
@@ -251,6 +253,53 @@ describe('Icon broadcast service', () => {
...
+      const references = createIconDefaultMap<string, IconMeta>();
+
+      const evt1 = [];
+      const broadcast1 = new IconsStateBroadcast(collections, references);
+      (broadcast1 as any).channel.addEventListener(
+        'message',
+        (e: MessageEvent<BroadcastIconsChangeMessage>) => {
+          evt1.push(e);
+        }
+      );
```
damyanpetev, line 268: Eh, not sure all of the extra broadcasts need to collect messages, the `channel` setup already does and will collect all other messages in `events` so you can check that collection instead of each extra broadcast :)

src/components/icon/icon.spec.ts
MayaKirova: But that doesn't make sense, does it? It's 3 new broadcast services, not 2. If we add the global one, it should make 5.

damyanpetev: It does, since the channel doesn't get its own messages - so with 3 services, each should get the replies from the 2 others + the initial request - so 3 in total :)

src/components/icon/icon.spec.ts
```diff
@@ -251,6 +253,53 @@ describe('Icon broadcast service', () => {
...
+      (broadcast3 as any).channel.addEventListener(
+        'message',
+        (e: MessageEvent<BroadcastIconsChangeMessage>) => {
+          evt3.push(e);
+        }
+      );
+
+      // a peer is requesting a state sync
+      channel.postMessage({ actionType: ActionType.SyncState });
+      await nextFrame();
```
damyanpetev, line 290: Also this seems to be flaky depending on the frame timing the messages are not always received. Sadly this:
```suggestion
      await aTimeout(20);
```
seems to be way more stable.

damyanpetev (approved):

rkaraivanov (approved):

dkamburov (approved):

PR #1536, opened 2025-01-13
refactor(dialog): provide message slot

### LINKED ISSUE #1509: Dialog content max-width

Dialog

### Description

The Dialog's content in webcomponents doesn't have any limitations in the width of the message. In angular it has a min-width of certain `ch` (characters) set for the different themes. In the UI kit the width of the container is set to "fixed" dimensions

Screenshot 2024-12-06 at 10 05 10
Screenshot 2024-12-06 at 10 05 17

### Reproduction

1. Open Dialog Story
2. Open Default Dialog

### Workaround

I have not found a workaround

### Is this a regression?

### Affected versions

all

all

SisIvanova:
Closes #1509


rkaraivanov (changes requested):
https://github.com/IgniteUI/igniteui-webcomponents/blob/02b18072ad86ef392878548d90a67f7dd859f721/src/components/dialog/dialog.ts#L38-L40

1. Please add a new entry for the message slot in the dialog API section.
2. If you have not updated the specification, please do so.
3. Add to the changelog

simeonoff (approved):

PR #1561, opened 2025-01-29
fix(rating): Value rounding calculation when step is not 1

### LINKED ISSUE #1560: [Bug]: Incorrect value rounding calculation when step is not 1 in igc-rating

Rating

### Description

When the rating's `step` is set to a value other than `1`, the rating value does not round correctly to the next valid step multiple. The current logic only rounds up, but it does not ensure that the resulting value is a multiple of the step.

### Reproduction

[CodeSandbox](https://codesandbox.io/p/sandbox/vibrant-sid-thqfkn) sample

1. Set up a rating component with a step other than 1 (e.g., 0.5 or 0.2).
2. Click on the component to select a value.
3. Observe that the selected value is rounded up but does not match a valid step multiple.

Image

### Workaround

I have a workaround by updating the rounding logic like this:
```
protected round(value: number) {
    return roundPrecise(
      Math.ceil(value / this.step) * this.step,
      numberOfDecimals(this.step)
    );
  }
```

### Is this a regression?

### Affected versions

Failing in 5.2.1, but worked in 5.0.0

Browser name = Chrome 131.0.0.0

Zneeky:
Closes #1560

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Fixed
+- #### Rating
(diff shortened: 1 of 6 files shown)


rkaraivanov (approved):
LGTM

Issue #1583, opened 2025-02-24
[Bug]: Chip long suffix and prefix wrap on 2 lines


simeonoff:
### Which component(s) are affected?

Chip

### Description

If the chip suffix or prefix contains long text, it wraps onto two lines. Since the chip has a fixed height, this causes a broken layout. A better approach is to apply an ellipsis to maintain visual consistency.

Originally posted by @simeonoff in https://github.com/IgniteUI/igniteui-angular/pull/14647#discussion_r1965060459

### Reproduction

See comment

### Workaround

I have not found a workaround.

### Is this a regression?

No or unsure. This never worker, or I haven't tried before.

### Affected versions

All

### Browser/OS/Node environment

All

PR #1588, opened 2025-02-26
fix(chip) styles for improved layout and responsiveness.

desig9stein:
Closes: #1583

Replaces flex with grid to mathc angular implementation and to make ellipsis work. Simplifies margin logic by using `gap` and removes redundant parts' styles. Adds `ellipsis()` to truncate text and sets a max width for slotted content to match angular behavior.

### DIFF
--- a/src/components/chip/themes/chip.base.scss
+++ b/src/components/chip/themes/chip.base.scss
@@ -2,21 +2,25 @@
 @use 'styles/utilities' as *;
 
 :host {
-    display: inline-flex;
+    display: inline-block;
     vertical-align: top;
+    height: ;
 }
 
 :host button {
     @include type-style('body-2');
+    @include ellipsis();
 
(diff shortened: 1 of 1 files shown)


src/components/chip/themes/chip.base.scss
```diff
@@ -2,21 +2,24 @@
...
+    display: grid;
+    grid-auto-flow: column;
     justify-content: center;
     align-items: center;
     font-family: var(--ig-font-family);
     border-style: none;
     box-shadow: none;
     cursor: pointer;
-    padding: pad(2px, 6px, 12px);
+    padding: 0 pad-inline(2px, 6px, 12px);
```
simeonoff, line 21:
```suggestion
    padding: 0 pad-inline(rem(2px), rem(6px), rem(12px));
```

PR #1608, opened 2025-03-17
Swap tiles based on moving direction

### LINKED ISSUE #1607: [Bug]: Multiple tile swapping when dragging tiles with different sizes

Tile Manager

### Description

When dragging a small tile over a bigger one, the tiles change their positions multiple times rearranging the layout.

Image

### Reproduction

Steps to reproduce:
1. Run the Fin Dashboard sample from Storybook.
2. Slowly drag the tile from the top left to the right (over "Order Line Items" tile).
3. Observe that the tiles swap multiple times.

Actual behavior:
The tiles swap multiple times.

Expected result:
The tiles should swap only once.

### Workaround

Swapping tiles with similar size or moving the pointer faster does not trigger this behavior

### Is this a regression?

No

### Affected versions

Reproducible in 5.3.0

N/A

mtsvyatkova:
Closes #1607


mtsvyatkova:
This scenario is added in the Tile Manager specification for manual testing: https://github.com/IgniteUI/igniteui-webcomponents/wiki/Tile-Manager-Specification#manual

src/components/tile-manager/tile.ts
```diff
@@ -402,14 +402,40 @@ export default class IgcTileComponent extends EventEmitterMixin<
...
+    if (this._previousPointerPosition) {
+      const deltaX = clientX - this._previousPointerPosition.x;
+      const deltaY = clientY - this._previousPointerPosition.y;
+
+      const movingRight = deltaX > 0;
+      const movingLeft = deltaX < 0;
+      const movingDown = deltaY > 0;
+      const movingUp = deltaY < 0;
+
+      const isHorizontalMove = Math.abs(deltaX) > Math.abs(deltaY);
```
gedinakova, line 419: It's very edgy, but if they are equal?

src/components/tile-manager/tile.ts
```diff
@@ -402,14 +402,40 @@ export default class IgcTileComponent extends EventEmitterMixin<
...
+      const isVerticalMove = Math.abs(deltaY) > Math.abs(deltaX);
```
rkaraivanov, line 420: I think we should abstract this state into the drag controller itself. It can hold the previous pointer coordinates when a drag is initiated and return the _direction_ while dragging in its callback parameters. At least it will simplify the code inside the tile drag handler.

rkaraivanov (changes requested):

src/components/common/controllers/drag.ts
```diff
@@ -6,14 +6,20 @@ import type {
...
+  roundByDPR,
+} from '../util.js';
 
 type DragCallback = (parameters: DragCallbackParameters) => unknown;
 type DragCancelCallback = (state: DragState) => unknown;
 
 export type DragCallbackParameters = {
   event: PointerEvent;
   state: DragState;
+  direction: Direction | null;
```
rkaraivanov, line 22: IMO this should be a property of the `DragState` object

src/components/common/controllers/drag.ts
```diff
@@ -249,7 +287,15 @@ class DragController implements ReactiveController {
     this._createDragGhost();
     this._updatePosition(event);
 
-    const parameters = { event, state: this._stateParameters };
+    this._previousPointerPosition = { x: event.clientX, y: event.clientY };
+    this._direction = this._trackPointerMovement(event.clientX, event.clientY);
```
rkaraivanov, line 291: I think the direction should be computed in the pointermove handler

src/components/common/controllers/drag.ts
```diff
@@ -28,6 +34,13 @@ type DragState = State & {
   element: Element | null;
 };
 
+type Direction = {
+  isHorizontalMove: boolean;
+  movingDown: boolean;
+  movingToEndHorizontally: boolean;
+  movingToStartHorizontally: boolean;
+};
```
rkaraivanov, line 42: Can this be a union type? Something along the lines of `left` | `top` | 'top-left`?

src/components/common/controllers/drag.ts
```diff
@@ -267,16 +313,30 @@ class DragController implements ReactiveController {
...
     this._updateMatcher(event);
 
-    const parameters = { event, state: this._stateParameters };
+    this._direction = this._trackPointerMovement(event.clientX, event.clientY);
+
+    const parameters = {
+      event,
+      state: this._stateParameters,
+      direction: this._direction,
+    };
```
rkaraivanov, line 322: If `direction` is a property of the `DragState` you can shove those calculations in the `_stateParameters` getter and only update the previous pointer coordinates here

src/components/common/controllers/drag.ts
```diff
@@ -267,16 +313,30 @@ class DragController implements ReactiveController {
...
+
     this._options.move?.call(this._host, parameters);
 
     this._assignPosition(this._dragItem);
   }
 
   private _handlePointerEnd(event: PointerEvent): void {
+    this._previousPointerPosition = null;
```
rkaraivanov, line 330: Maybe just initialize the previous coordinates as `{x: 0, y: 0}` on initialization of the controller and then just updated through the pointerdown -> pointermove cycle?

src/components/common/controllers/drag.ts
```diff
@@ -109,6 +122,8 @@ class DragController implements ReactiveController {
   private _hasPointerCapture = false;
 
   private _ghost: HTMLElement | null = null;
+  private _previousPointerPosition: { x: number; y: number } | null = null;
+  private _direction: Direction | null = null;
```
rkaraivanov, line 126: Won't need it if it's a part of the `DragState` object

rkaraivanov (changes requested):

rkaraivanov (approved):

PR #1613, opened 2025-03-20
feat(file-input): Added file input component

### LINKED ISSUE #1581: [FR]: Support type="file" for File Upload
Currently, <igc-input> does not support the type="file" functionality for handling file uploads. Adding support for file input would allow users to select and upload files directly using Ignite UI for Web Components, aligning with native HTML <input type="file"> behavior.

Expected Behavior:
- The component should allow users to select files from their local system.
- Support for single and multiple file selection (multiple attribute).
- Ability to capture file metadata (name, size, type).
- Event(s) for handling file selection and retrieval.

This is the bear minimum I can think of, although we might need to comply with the Angular component first, lets discuss thi.

simeonoff:
Closes #1581


rkaraivanov (changes requested):
LGTM, with some minor style suggestions. 

The main concern is a UX issue with required file inputs. When a user clicks a required file input, the focus shifts to the browser's file dialog. This triggers the input's blur event, which in turn runs the validation logic. This can lead to immediate validation errors being displayed even before the user has selected a file, which is not ideal. To fix this, please implement an internal boolean flag (e.g., `_hasActivation`).
Set this flag to true on the input's click event (activation), and reset it to false on the change or cancel events. Then, modify the blur handler to skip validation if `_hasActivation` is true.

rkaraivanov (approved):

SisIvanova (approved):

Issue #1596, opened 2025-03-06
Date Range Picker Component


kacheshmarova:
As a user I want to have Date Range Picker Component in WC that has the same design, UX and features as Angular Tooltip.

Estimated Delivery Sprint: Sprint 5, 30 April

- [x] Research
- [x] Specification
- [x] Component and tests implementation
- [x] API Documentation
- [x] Design implementation

PR #1614, opened 2025-03-21
Adding Date Range Picker Component

IvanKitanov17:
Closes: [#1596](https://github.com/IgniteUI/igniteui-webcomponents/issues/1596)


src/components/date-time-input/date-time-input.ts
```diff
@@ -123,6 +123,25 @@ export default class IgcDateTimeInputComponent extends EventEmitterMixin<
...
+
+  @property()
+  public set displayDateRange(value: (Date | null)[]) {
+    this._displayDateRange = value;
+    this.setDateRangePlaceHolder();
+    if (value && value.length === 2) {
+      this.readOnly = true;
+      this.updateMask();
+    }
+  }
```
rkaraivanov, line 143: Not a fan of a sprouting a new property on a public component...

src/components/common/definitions/defineAllComponents.ts
```diff
@@ -1,3 +1,4 @@
+import { IgcDateRangePickerComponent } from '../../../index.js';
```
rkaraivanov, line 1: Don't import from the top barrel file

rkaraivanov (commented):
Just a quick glance over the current implementation...

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -1013,7 +1013,9 @@ export default class IgcDateRangePickerComponent extends FormAssociatedRequiredM
         slot="prefix"
         part=${calendarIcon}
         @pointerdown=${this._handleCalendarIconSlotPointerDown}
-        @click=${this.handleAnchorClick}
+        @click=${() => {
```
rkaraivanov, line 1016:
```suggestion
        @click=${this.readOnly ? nothing : this.handleAnchorClick}
```
As an added bonus this will detach the existing event listener

src/components/date-range-picker/predefined-ranges-area.ts
```diff
@@ -0,0 +1,119 @@
...
+   */
+  @property({ type: Array })
+  public customRanges: CustomDateRange[] = [];
+
+  /** The resource strings of the date range area component. */
+  @property({ attribute: false })
+  public resourceStrings: IgcDateRangePickerResourceStrings =
+    IgcDateRangePickerResourceStringsEN;
+
+  private _predefinedRanges: CustomDateRange[] = [
```
rkaraivanov, line 49: I see that you have tests which cover setting a specific localization property
and verifying it shows correctly here. That is fine as long as the `resourceStrings` reference is the same. Have you tested what happens when a developer passes a whole new resource object to the date range picker and how that reflects in the predefined ranges?

src/components/date-range-picker/predefined-ranges-area.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.dispatchEvent(
+      new CustomEvent('rangeSelect', { detail: range.dateRange })
+    );
+  }
+
+  protected override render() {
+    return html`
+      <div
+        class="ranges"
+        style="display: flex; flex-direction: row; gap: 5px; margin-top: 20px;"
```
rkaraivanov, line 101: Should be abstracted in the styles but it will be probably...

src/components/date-range-picker/predefined-ranges-area.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  private get _allRanges(): CustomDateRange[] {
+    return this.usePredefinedRanges
+      ? [...this._predefinedRanges, ...this.customRanges]
+      : [...this.customRanges];
+  }
+
+  private _handleRangeSelect(range: CustomDateRange) {
+    this.dispatchEvent(
+      new CustomEvent('rangeSelect', { detail: range.dateRange })
```
rkaraivanov, line 93: Maybe prefix the event since this seems kind of generic. It doesn't bubble up and it is not composed but still 😄

src/components/calendar/helpers.ts
```diff
@@ -71,6 +76,33 @@ export function convertToDate(value?: Date | string | null): Date | null {
...
+  value?: DateRangeValue | string | null
+): DateRangeValue | null {
+  if (!value) {
+    return null;
+  }
+
+  if (isString(value)) {
+    const obj = JSON.parse(value);
+    const start = isValidDate(new Date(obj.start));
+    const end = isValidDate(new Date(obj.end));
```
rkaraivanov, line 97:
```suggestion
    const start = convertToDate(obj.start);
    const end = convertToDate(obj.end);
```
I would like to keep accepting only valid ISO 8601 formatted strings.

src/components/calendar/helpers.ts
```diff
@@ -81,6 +113,33 @@ export function getDateFormValue(value: Date | null) {
...
+  value: DateRangeValue | null,
+  host: IgcFormControl
+): FormValueType {
+  if (!value?.start && !value?.end) return null;
+
+  const start = value?.start?.toISOString();
+  const end = value?.end?.toISOString();
+
+  const fd = new FormData();
+  const prefix = host.name ? `${host.name}-` : '';
```
rkaraivanov, line 135: If the date range picker does not have a set **name**, then it should **not** participate in the native form submit. Also, what is the behavior in the case where just one part of the range is set? The component skips setting the other at all?

src/components/calendar/helpers.ts
```diff
@@ -81,6 +113,33 @@ export function getDateFormValue(value: Date | null) {
...
+/**
+ * Converts a DateDateRangeValue object to FormData with
+ * start and end Date values as ISO 8601 strings.
+ * The keys are prefixed with the host name if it exists
+ * and suffixed with 'start' or 'end' accordingly.
+ *
+ * If the date values are null or undefined, the form data values
+ * are empty strings ''.
+ */
+export function getDateRangeFormValue(
```
rkaraivanov, line 125: I would move this function in the `form-value.ts` file since it's specific to that behavior.

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+  igcOpened: CustomEvent<void>;
+  igcClosing: CustomEvent<void>;
+  igcClosed: CustomEvent<void>;
+  igcChange: CustomEvent<DateRangeValue | null>;
+  igcInput: CustomEvent<DateRangeValue | null>;
+}
+
+// #endregion
+
+// #region Annotations
```
rkaraivanov, line 94: Yeah, that's an overkill. Use the `// #region` stuff only for code, not documentation and metadata.

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+    this._updateMaskedRangeValue();
+  }
+
+  public get value(): DateRangeValue | null {
+    return this._formValue.value;
+  }
+  /**
+   * Renders chips with custom ranges based on the elements of the array.
+   */
+  @property({ type: Array })
```
rkaraivanov, line 301: I'm on the fence, whether this should be an attribute

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+   */
+  @property()
+  public prompt = '_';
+
+  /**
+   * Format to display the value in when not editing.
+   * Defaults to the input format if not set.
+   * @attr display-format
+   */
+  @property({ attribute: 'display-format', reflect: false })
```
rkaraivanov, line 424:
```suggestion
  @property({ attribute: 'display-format' })
```
Lit does not reflect attributes by default

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+
+  // #endregion
+
+  // #region Calendar properties
+
+  /**
+   * The number of months displayed in the calendar.
+   * @attr visible-months
+   */
+  //
```
rkaraivanov, line 503:
```suggestion
```

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+  @property({ attribute: false })
+  public specialDates!: DateRangeDescriptor[];
+
+  /** Sets the start day of the week for the calendar. */
+  @property({ attribute: 'week-start' })
+  public weekStart: WeekDays = 'sunday';
+
+  // #endregion
+
+  // #endregion
```
rkaraivanov, line 570: Duplicated?

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+  @watch('useTwoInputs')
+  protected async _updateDateRange() {
+    await this._calendar?.updateComplete;
+    this._updateMaskedRangeValue();
+    this._setCalendarRangeValues();
+  }
+
+  @watch('mode')
+  protected async _modeChanged() {
+    if (this.mode === 'dialog') {
```
rkaraivanov, line 655:
```suggestion
    if (!this._isDropDown) {
```
I see there are a couple of checks and you already have a utility to check it

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+    if (!this._calendar) {
+      return;
+    }
+
+    if (!(this.value?.start || this.value?.end)) {
+      this._calendar.values = null;
+      return;
+    }
+
+    if (!(this.value?.start && this.value?.end)) {
```
rkaraivanov, line 893: I see a lot of checks for `value.start` and `value.end`. Can you abstract it in a helper getter/method?

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+  // Checks for dirty state to avoid unnecessary validation on form reset,
+  // caused by the inputs value being set.
+  private _delegateInputsValidity() {
+    const inputs = this.useTwoInputs ? this._inputs : [this._input];
+
+    inputs.forEach((input) => {
+      input.checkValidity = () =>
+        this._dirty || !this._pristine ? this.checkValidity() : true;
+      input.reportValidity = () =>
+        this._dirty || !this._pristine ? this.reportValidity() : true;
```
rkaraivanov, line 822: ?! Why the need for additional flag checking?

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -0,0 +1,1211 @@
...
+
+    if (start.equalTo(end)) {
+      this._calendar.values = [this.value.start!];
+    } else {
+      const range = Array.from(
+        calendarRange({ start: this.value.start!, end: this.value.end! })
+      );
+      // calendarRange is non-inclusive
+      range.push(last(range).add('day', start.lessThan(end) ? 1 : -1));
+      this._calendar.values = range.map((d) => d.native);
```
rkaraivanov, line 910: You can extend the `calendarRange` method to support inclusive range with config parameter, right 😉 ? Better than working around it

rkaraivanov (changes requested):
LGTM with some minor issues.
I have yet to go over the validator stuff

yradoeva:
**All themes, Two inputs, Dialog mode** - from interaction point of view, the behaviour is correct. But visually, the inputs look as readonly. That “readonly” is incorrectly written in the design handoff (will be fixed), the more appropriate state for the inputs would be “non-editable”.
_Clicking anywhere in the input (incl. toggle icons) opens the calendar picker in a dialog. Clicking on the clear icon/s clears the selected dates._
Both inputs should look visually as the examples with single input.

Implementation:
image
Design:
image

yradoeva:
**All themes, Two inputs:** Input width should not change when entering/clearing value. Width should be the same in idle and filled.

https://github.com/user-attachments/assets/fe814bd6-efc7-4453-b4e5-64955c65a6cb

yradoeva:
**Material, Outlined (Border) type** - there is a slight glitch when you click on the toggle icon:

https://github.com/user-attachments/assets/8b82ccf9-70d8-4e7d-b930-3a6ccf1900f7

yradoeva:
**All themes, Readonly mode (single or 2 inputs) with Dialog:** the dialog should not open on click in the input, as it doesn't open on click on the toggle. 

https://github.com/user-attachments/assets/e8f9d88b-c27b-4121-9d3c-359bdcdda217

yradoeva:
[Input] [Readonly] [Fluent] The suffix icon color for Input type: Search, readonly focused should be light grays.500 / dark grays.300. 
image

src/components/date-range-picker/date-range-picker.ts
```diff
@@ -286,7 +286,7 @@ export default class IgcDateRangePickerComponent extends FormAssociatedRequiredM
   /**
    * Renders chips with custom ranges based on the elements of the array.
    */
-  @property({ type: Array, attribute: 'custom-ranges' })
+  @property({ type: Array, attribute: false })
```
rkaraivanov: You can drop the `type: Array` as well

rkaraivanov (approved):

PR #1629, opened 2025-03-31
[Calendar] - fix header size

### LINKED ISSUE #1609: [Calendar]: Vertical header text gets wrapped to three lines
When we set our calendar to a vertical position, the text in the header gets wrapped into three lines. In the UI kits, the text appears on only two lines, as in the Angular sample. The issue occurs when using the Material, Bootstrap, and Indigo themes, in both light and dark versions.

Material:

Image

Bootstrap:

Image

Indigo:

Image
--------

I found that the difference between the Angular and WC samples is that the WC sample is missing the CSS property 'min-width,' which is equal to the 'width' CSS property on the same element. After adding it, the issue seems to be fixed.

Image

desig9stein:
closes #1609

This PR Also fixes the bootstrap first-row border position when the calendar is in multiple calendar view
Screenshot 2025-03-31 at 13 18 08


didimmova (commented):
Can't you add the header size in the base as they're all having the same size, and just override it for indigo?

didimmova (approved):

PR #1637, opened 2025-04-02
fix(calendar): address ARIA violation when a week with all days hidden is rendered

### LINKED ISSUE #1636: [Bug] Calendar throws ARIA violations in certain configurations

Calendar

### Description

When `visibleMonths` is set to at least 2, for certain combinations of days-views, such as April+May, depending on the number of days, sometimes an empty "week" is rendered, where all days are _hidden_. This leads to a `Certain ARIA roles must contain particular children` violation.

Image

### Reproduction

1. Set visibleMonths to 2
2. Go to April 2025

### Workaround

In case all days in a week are hidden, the week should have `aria-hidden` set to `true`, or not be rendered at all?

### Is this a regression?

### Affected versions

5.3.0

Chrome, Mozilla, etc.

ddaribo:
Closes #1636 

Note: An alternative approach might be to conditionally yield nothing in the `renderWeeks` generator if all days are hidden.


src/components/calendar/days-view/days-view.ts
```diff
@@ -398,11 +398,10 @@ export default class IgcDaysViewComponent extends EventEmitterMixin<
     };
   }
 
-  protected renderDay(day: CalendarDay, today: CalendarDay) {
+  protected renderDay(day: CalendarDay, props: any) {
```
rkaraivanov, line 401: I would prefer to not change the signature of this method.

src/components/calendar/days-view/days-view.ts
```diff
@@ -481,12 +480,17 @@ export default class IgcDaysViewComponent extends EventEmitterMixin<
     const last = weeks.length - 1;
 
     for (const [idx, week] of weeks.entries()) {
+      const weekDaysProps: any[] = [];
+      for (const day of week) {
+        weekDaysProps.push(this.getDayProperties(day, today));
+      }
+      const hidden = weekDaysProps.every((p) => p.hidden);
```
rkaraivanov, line 487: IMO it would be better to check just the first and the last week whether all the days are hidden before yielding the DOM pieces.

rkaraivanov (changes requested):
Probably yielding nothing would be the better approach but that can be a separate refactoring task.

src/components/calendar/days-view/days-view.ts
```diff
@@ -481,8 +482,14 @@ export default class IgcDaysViewComponent extends EventEmitterMixin<
...
     for (const [idx, week] of weeks.entries()) {
+      let hidden: boolean | undefined;
+      if (idx === 0 || idx === last) {
+        hidden =
+          week.every((day) => this.getDayProperties(day, today).hidden) ||
+          undefined;
+      }
       yield html`
-        <div role="row" part="days-row">
+        <div role="row" part="days-row" aria-hidden=${ifDefined(hidden)}>
```
rkaraivanov, line 492: No point going the undefined route for the aria-hidden attribute.
Something along the lines of:
```ts
for (const [idx, week] of weeks.entries()) {
      let hidden = false;
      if (idx === 0 || idx === last) {
        hidden = week.every((day) => this.getDayProperties(day, today).hidden);
      }

      yield html`
        <div role="row" part="days-row" aria-hidden=${hidden}>
          ${this.showWeekNumbers
            ? this.renderWeekNumber(week[0], idx === last)
            : nothing}
          ${week.map((day) => this.renderDay(day, today))}
        </div>
      `;
    }
```

rkaraivanov (changes requested):

rkaraivanov (approved):

PR #1638, opened 2025-04-02
fix(forms): Setting validation properties on a pristine non-dirty form

### LINKED ISSUE #1632: [Bug]: Input required state is wrongly set, even though component is prestine

Input

### Description

When an `igc-input` required state is set dynamically to `true`, then the state of the component is set to required, even when the input is pristine. This also affects the react input, where the required state is set dynamically. (see sample: [validation_react.zip](https://github.com/user-attachments/files/19541071/validation_react.zip))

### Reproduction

1. Open the following sample: https://igniteui.github.io/igniteui-webcomponents/?path=/story/input--form
2. Inspect the first input, which is not required, and force it to the `:active` state.
3. Through the developer console, set dynamically the required state of the input to true: `$0.required = true` 
4. Even though the component is pristine, the required state of the component is required state:

Image

### Workaround

I have not found a workaround

### Is this a regression?

### Affected versions

4.3.x

Browser: Chrome

rkaraivanov:
Setting validation properties on a pristine non-dirty form element does not apply invalid styles.

Closes #1632


damyanpetev (approved):

Marina-L-Stoyanova:
I have tested the fix with the attached sample as well as the sample in codegen-output repo : https://github.com/IgniteUI/codegen-output/tree/master/04_Forms/Validation%20Samples/React%20validation and it is working as expected.

PR #1651, opened 2025-04-09
fix(input-group): fluent theme discrepancies

SisIvanova:
Related to IgniteUI/igniteui-theming/issues/393


didimmova (dismissed):

didimmova (approved):

didimmova (dismissed):

yradoeva:
**Date Picker - focused / focused & hover**
Label color should be the same as in idle state —idle-secondary-color (for light and dark mode), currently it gets
host(focus-within)  part=label  {
which is not correct for light mode, should be light grays.900

didimmova (approved):

PR #1684, opened 2025-05-08
refactor: Keyboard focus ring controller

rkaraivanov:


src/components/carousel/carousel.ts
```diff
@@ -328,6 +327,13 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
 
     this.addEventListener('pointerenter', this.handlePointerEnter);
     this.addEventListener('pointerleave', this.handlePointerLeave);
+    this.addEventListener('pointerdown', () => {
+      this._hasInnerFocus = false;
+    });
+
+    this.addEventListener('keyup', () => {
+      this._hasInnerFocus = true;
+    });
```
damyanpetev, line 336: Shouldn't these be replaced with just `this._hasInnerFocus = true;` on the `focusin` handler?

damyanpetev: Okay, I get that's the current behavior replicating what the focus ring did, but it doesn't make sense. 
See https://igniteui.github.io/igniteui-webcomponents/?path=/story/carousel--inputs-template&args=interval:3000
Where focusing an input and moving the pointer away starts changing slides, which I don't think is nice and we should consider updating that behavior.

damyanpetev (approved):
Outside of the current behavoir comment and we can address separately, the actual refactor LGTM

PR #1688, opened 2025-05-14
feat(chat): add chat component

igdmdimitrov:


src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,187 @@
...
+import { styles } from './themes/input.base.css.js';
+import {
+  type IgcMessageAttachment,
+  attachmentIcon,
+  sendButtonIcon,
+} from './types.js';
+
+/**
+ *
+ * @element igc-chat
```
gedinakova, line 19: @teodosiah This change would stop the storybook confusion when rebuilding.
```suggestion
 * @element igc-chat-input
```

src/components/chat/types.ts
```diff
@@ -0,0 +1,227 @@
...
+  currentUserId?: string;
+  /**
+   * If `true`, prevents the chat from automatically scrolling to the latest message.
+   */
+  disableAutoScroll?: boolean;
+  /**
+   * If `true`, disables the ability to upload and send attachments.
+   * Defaults to `false`.
+   */
+  disableAttachments?: boolean;
```
igdmdimitrov: Since this is affecting the attachments option in the input area we should either rename it to something like `disableInputAttachments` or hide attachments in all messages.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+  private _activeMessageId = '';
+
+  /**
+   * Registers this component and its dependencies.
+   * Used internally for component setup.
+   */
+  /* blazorSuppress */
+  public static register() {
+    registerComponent(IgcChatMessageListComponent, IgcChatMessageComponent);
+  }
```
rkaraivanov: Move it in the beginning of the class with other static members.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+   */
+  @consume({ context: chatContext, subscribe: true })
+  private _chatState?: ChatState;
+
+  /**
+   * Currently active message id used for focus and keyboard navigation.
+   * @private
+   */
+  @state()
+  private _activeMessageId = '';
```
rkaraivanov: Since we won't do a custom keyboard navigation this can be dropped.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  constructor() {
+    super();
+    addKeybindings(this)
+      .set(homeKey, this.navigateToMessage)
+      .set(endKey, this.navigateToMessage)
+      .set(arrowUp, this.navigateToMessage)
+      .set(arrowDown, this.navigateToMessage);
+  }
```
rkaraivanov: No keyboard navigation - up for removal

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+  private scrollToBottom() {
+    requestAnimationFrame(() => {
+      const container = this.shadowRoot?.host as HTMLElement;
+      if (container) {
+        container.scrollTop = container.scrollHeight;
+      }
+    });
+  }
+
+  /**
```
rkaraivanov: Again, since initially there won't be any custom keyboard navigation implementation, it is up for removal.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+   * Scrolls the view to a specific message by id.
+   * @param messageId - The id of the message to scroll to
+   * @private
+   */
+  private scrollToMessage(messageId: string) {
+    const messageElement = this.shadowRoot?.querySelector(
+      `#message-${messageId}`
+    );
+    messageElement?.scrollIntoView();
+  }
```
rkaraivanov: This is potentially useful as an API on the chat component itself.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  /**
+   * Lifecycle: called after first render.
+   * Scrolls to bottom unless auto-scroll is disabled.
+   */
+  protected override firstUpdated() {
+    if (!this._chatState?.options?.disableAutoScroll) {
+      this.scrollToBottom();
+    }
```
rkaraivanov: There is no guarantee that the child elements of the list will be completely rendered and updated at this point, in the browser. Maybe that's why the `scrollToBottom` is wrapped inside a rAF?

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+        grouped[dateStr] = [];
+      }
+      grouped[dateStr].push(message);
+    });
+
+    return Object.keys(grouped).map((date) => ({
+      date,
+      messages: grouped[date],
+    }));
+  }
```
rkaraivanov: I can't find in the specification where we've agreed to group by timestamps as a default behavior. It seems like that is up to the developer to do and we should just provide an API if such is missing.

src/components/chat/chat-message-list.ts
```diff
@@ -0,0 +1,315 @@
...
+    if (date.toDateString() === yesterday.toDateString()) {
+      return 'Yesterday';
+    }
+
+    return date.toLocaleDateString('en-US', {
+      weekday: 'long',
+      month: 'short',
+      day: 'numeric',
+    });
+  }
```
rkaraivanov: Again, as config option in the chat API. We could have a default formatter out of the box, but hardcoding it here means it cannot be changed by the developer.

src/components/chat/chat-message.ts
```diff
@@ -0,0 +1,123 @@
...
+   * This is used internally to set up the component definitions.
+   */
+  /* blazorSuppress */
+  public static register() {
+    registerComponent(
+      IgcChatMessageComponent,
+      IgcMessageAttachmentsComponent,
+      IgcAvatarComponent
+    );
+  }
```
rkaraivanov: At the start of the class with other static members.

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+    if (fileType?.startsWith('text')) {
+      return 'file-document';
+    }
+    if (fileType?.startsWith('image')) {
+      return 'file-image';
+    }
+    return 'file-other';
+  }
+
+  public get defaultAttachmentsArea(): TemplateResult {
```
rkaraivanov: First of all, these `default*` should be actual methods and no getters.
Then since they return actual DOM templates, they are renderers, so I expect them to start
with `render*`.
Finally, why are they public?

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  public get defaultTextArea(): TemplateResult {
+    return html` <igc-textarea
+      part="text-input"
+      .placeholder=${this.inputPlaceholder}
+      resize="auto"
+      rows="1"
+      .value=${this.inputValue}
+      @input=${this.handleInput}
```
rkaraivanov: Please use the igcInput event

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+    this.inputValue = this._chatState?.inputValue || '';
+    this.inputPlaceholder = this._chatState?.options?.inputPlaceholder || '';
+  }
+
+  private handleInput(e: Event) {
+    const target = e.target as HTMLTextAreaElement;
+    this.inputValue = target.value;
+    this._chatState?.handleInputChange(this.inputValue);
+    // this.adjustTextareaHeight();
+  }
```
rkaraivanov:
```suggestion
  private handleInput({ detial }: CustomEvent<string>) {
    this.inputValue = detail;
    this._chatState?.handleInputChange(this.inputValue);
    // this.adjustTextareaHeight();
  }
```
If you bind to the igcInput handler.

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  private handleKeyDown(e: KeyboardEvent) {
+    if (e.key === 'Enter' && !e.shiftKey) {
+      e.preventDefault();
+      this.sendMessage();
+    } else {
```
rkaraivanov: There is a keybinding controller which abstracts these checks for you

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+      this._chatState?.emitEvent('igcTypingChange', {
+        detail: { isTyping: true },
+      });
+
+      // wait 3 seconds and dispatch a stop-typing event
+      setTimeout(() => {
+        this._chatState?.emitEvent('igcTypingChange', {
+          detail: { isTyping: false },
+        });
+      }, 3000);
```
rkaraivanov: I don't like the hardcoded 3s timeout. It should be exposed as a config option

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+
+  protected *renderDefaultFileUploadTemplate() {
+    yield html`${this._chatState?.options?.disableAttachments
+      ? nothing
+      : this.defaultFileUploadButton}`;
+  }
+
+  protected override firstUpdated() {
+    this.setupDragAndDrop();
+    if (this._chatState) {
```
rkaraivanov: In what scenario is the state object missing?

src/components/chat/chat-input.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  protected *renderDefaultFileUploadTemplate() {
+    yield html`${this._chatState?.options?.disableAttachments
+      ? nothing
+      : this.defaultFileUploadButton}`;
+  }
+
+  protected override firstUpdated() {
+    this.setupDragAndDrop();
```
rkaraivanov: Maybe move this in the connected callback?
Also, I don't see where the drag handlers are cleaned up?

src/components/chat/types.ts
```diff
@@ -0,0 +1,231 @@
...
+export const fileImageIcon =
+  '<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#e3e3e3"><path d="M200-120q-33 0-56.5-23.5T120-200v-560q0-33 23.5-56.5T200-840h280v80H200v560h560v-280h80v280q0 33-23.5 56.5T760-120H200Zm188-212-56-56 372-372H560v-80h280v280h-80v-144L388-332Z"/></svg>';
+export const fileDocumentIcon =
+  '<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#e3e3e3"><path d="M200-120q-33 0-56.5-23.5T120-200v-560q0-33 23.5-56.5T200-840h280v80H200v560h560v-280h80v280q0 33-23.5 56.5T760-120H200Zm188-212-56-56 372-372H560v-80h280v280h-80v-144L388-332Z"/></svg>';
+export const fileOtherIcon =
+  '<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#e3e3e3"><path d="M200-120q-33 0-56.5-23.5T120-200v-560q0-33 23.5-56.5T200-840h280v80H200v560h560v-280h80v280q0 33-23.5 56.5T760-120H200Zm188-212-56-56 372-372H560v-80h280v280h-80v-144L388-332Z"/></svg>';
```
rkaraivanov: These should probably go in the icon registry with their respective configuration.
@simeonoff

src/components/chat/message-attachments.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (attachment.url) {
+      return attachment.url;
+    }
+    if (attachment.file) {
+      return URL.createObjectURL(attachment.file);
+    }
+    return '';
+  }
+
+  private defaulAttachmentContent(attachment: IgcMessageAttachment) {
```
rkaraivanov:
```suggestion
  private renderDefaulAttachmentContent(attachment: IgcMessageAttachment) {
```

src/components/chat/chat.ts
```diff
@@ -0,0 +1,335 @@
...
+          `
+        )}
+      </slot>
+      <slot name="suggestions-actions" part="suggestions-actions"></slot>
+    </div>`;
+  }
+
+  protected override firstUpdated() {
+    this._context.setValue(this._chatState, true);
+  }
```
rkaraivanov: Not needed since consumers (other chat component UI components) will query the parent chat

src/components/chat/chat-state.ts
```diff
@@ -0,0 +1,354 @@
...
+  /** The current list of messages */
+  private _messages: IgcMessage[] = [];
+  /** Sorted copy of messages by timestamp */
+  private _sortedMessages: IgcMessage[] = [];
+  /** Chat options/configuration */
+  private _options?: IgcChatOptions;
+  /** List of current input attachments */
+  private _inputAttachments: IgcMessageAttachment[] = [];
+  /** Current input text */
+  private _inputValue = '';
```
rkaraivanov: Formatting

src/components/chat/chat-state.ts
```diff
@@ -0,0 +1,354 @@
...
+    'below-messages';
+  //#endregion
+
+  //#region Public properties
+  /**
+   * Array of message IDs sorted by timestamp ascending.
+   */
+  public get sortedMessagesIds(): string[] {
+    return this._sortedMessages.map((m) => m.id);
+  }
```
rkaraivanov: Was that used for the keyboard navigation implementation?

src/components/chat/chat-state.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  /**
+   * Sets the list of chat messages.
+   * Also sorts messages by timestamp and requests host update.
+   */
+  public set messages(value: IgcMessage[]) {
+    this._messages = value;
+    this._sortedMessages = value.slice().sort((a, b) => {
+      return a.timestamp.getTime() - b.timestamp.getTime();
+    });
```
rkaraivanov: Not a fan, similar to the group by behavior in the list component.
Up to the developer to sort them, or we expose an API hook for the user to provide a sort comparator.

src/components/chat/chat.ts
```diff
@@ -0,0 +1,335 @@
...
+  protected override createRenderRoot(): HTMLElement | DocumentFragment {
+    const root = super.createRenderRoot();
+    root.addEventListener('slotchange', () => {
+      this._hasContent =
+        this._chatState?.hasSlotContent('prefix') ||
+        this._chatState?.hasSlotContent('title') ||
+        this._chatState?.hasSlotContent('actions');
+    });
+    return root;
+  }
```
rkaraivanov: We do have a [slot controller](https://github.com/IgniteUI/igniteui-webcomponents/blob/master/src/components/common/controllers/slot.ts) for this kind of stuff.

src/components/chat/chat.ts
```diff
@@ -0,0 +1,335 @@
...
+  constructor() {
+    super();
+    addThemingController(this, all);
+  }
+
+  /**
+   * The list of chat messages currently displayed.
+   * Use this property to set or update the message history.
+   */
+  @property({ reflect: true, attribute: false })
```
rkaraivanov:
```suggestion
  @property({ attribute: false })
```

src/components/chat/chat.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  public get messages(): IgcMessage[] {
+    return this._chatState.messages;
+  }
+
+  /**
+   * The chat message currently being composed but not yet sent.
+   * Includes the draft text and any attachments.
+   */
+  @property({ reflect: true, attribute: false })
```
rkaraivanov:
```suggestion
  @property({ attribute: false })
```

src/components/chat/chat.ts
```diff
@@ -0,0 +1,335 @@
...
+  public set draftMessage(value: {
+    text: string;
+    attachments?: IgcMessageAttachment[];
+  }) {
+    if (this._chatState) {
+      this._chatState.inputValue = value.text;
+      this._chatState.inputAttachments = value.attachments || [];
+      this.requestUpdate();
+    }
+  }
```
rkaraivanov: What if a nullish value is passed in? Don't see any checks here?

src/components/chat/chat-input.ts
gedinakova: Is it necessary? They are added to the input-container which is in the shadow DOM of the input.

rkaraivanov (approved):

PR #1690, opened 2025-05-15
fix(textarea): Dynamic prefix/suffix slot content state

### LINKED ISSUE #1686: [igc-textarea]: When dynamically removing the prefix and suffix from the textarea input, their wrappers remain visible.
When we dynamically remove the prefix and suffix from the textarea input, only their content is removed, the div wrappers remain visible and take up space.

Image

The issue occurs on all themes.

rkaraivanov:
Closes #1686


rkaraivanov:
@simeonoff 
The actual fix is [here](https://github.com/IgniteUI/igniteui-webcomponents/blob/f8893cd14934fda6d463de3341d37c5149370d5c/src/components/textarea/textarea.ts#L358-L374) the rest is an internal renaming and cleanups.

Sorry for the changeset diff noise

PR #1692, opened 2025-05-16
refactor(date-picker): Interaction behavior in readonly mode

rkaraivanov:
* The same behavior should be applied to #1614 as well.
* Relevant specifications should be updated to reflect the new behavior.


src/components/date-picker/date-picker.ts
```diff
@@ -529,35 +542,30 @@ export default class IgcDatePickerComponent extends FormAssociatedRequiredMixin(
     event.preventDefault();
   }
```
ddaribo, line 543:
```ts
protected handlerCalendarIconSlotPointerDown(event: PointerEvent) {
``` 
->
```ts
protected _handleCalendarIconSlotPointerDown(event: PointerEvent) {
```
maybe rename this method as well?

PR #1694, opened 2025-05-16
feat(radio-group): group label styles

SisIvanova:
Related to IgniteUI/igniteui-angular/issues/15554

### DIFF
--- a/src/components/radio-group/radio-group.ts
+++ b/src/components/radio-group/radio-group.ts
@@ -1,14 +1,13 @@
 import { LitElement, html } from 'lit';
 import { property, queryAssignedElements } from 'lit/decorators.js';
-
 import { themes } from '../../theming/theming-decorator.js';
+import { createMutationController } from '../common/controllers/mutation-observer.js';
 import { registerComponent } from '../common/definitions/register.js';
 import IgcRadioComponent from '../radio/radio.js';
 import type { ContentOrientation } from '../types.js';
-import { styles } from './radio-group.base.css.js';
-import { styles as fluent } from './radio-group.fluent.css.js';
-import { styles as indigo } from './radio-group.indigo.css.js';
-import { styles as material } from './radio-group.material.css.js';
+import { styles } from './themes/radio-group.base.css.js';
+import { styles as shared } from './themes/shared/radio-group.common.css.js';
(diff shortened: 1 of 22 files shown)


src/components/radio-group/radio-group.ts
```diff
@@ -35,6 +31,12 @@ export default class IgcRadioGroupComponent extends LitElement {
   private _name!: string;
   private _value!: string;
 
+  @property({ type: Boolean, reflect: true })
+  private disabled = false;
+
+  @property({ type: Boolean, reflect: true })
+  private labelBefore = false;
```
rkaraivanov, line 38: If these are internal they should be `@state`

src/components/radio-group/radio-group.ts
```diff
@@ -90,11 +92,29 @@ export default class IgcRadioGroupComponent extends LitElement {
     return this._value;
   }
 
+  private _observerCallback() {
+    const radios = Array.from(this._radios as unknown as Element[]).filter(
+      (el) => el.tagName.toLowerCase() === 'igc-radio'
+    ) as IgcRadioComponent[];
```
rkaraivanov, line 98:
```suggestion
    const radios = Array.from(this._radios)
```

Isn't `this._radios` already typed as an `NodeListOf<IgcRadioComponent>`?

rkaraivanov (changes requested):

src/components/radio-group/radio-group.ts
```diff
@@ -93,12 +87,20 @@ export default class IgcRadioGroupComponent extends LitElement {
...
+      condition
+        ? this._internals.states.add(state)
+        : this._internals.states.delete(state);
+
+    setState(
+      'disabled',
+      radios.every((radio) => radio.disabled)
+    );
+    setState(
+      'labelBefore',
```
rkaraivanov, line 101:
```suggestion
      'label-before',
```
IMO it would be better to stick to the native CSS pseudo selectors/elements schema, which is usually kebab cased; `nth-child`, `first-of-type`, etc.

rkaraivanov (changes requested):

simeonoff (approved):

rkaraivanov (approved):

PR #1696, opened 2025-05-16
fix(date-time-input): ensure 'igcChange' is emitted in case an incomplete mask value has been filled

### LINKED ISSUE #1695: [Bug]: IgcDateTimeInput does not emit `igcChange` when a mask containing empty character has been filled

Date picker, Date-time input

### Description

IgcDateTimeInput does not emit `igcChange` when a mask containing empty character has been filled.
Components such as the `date-picker` and the new `date-range-picker` listen for the event to reflect that a value has been set.

In the below gif, it can be seen that because the event is not emitted, the new value is not reflected and the clear icon does not appear, although the input now has value:

Image

### Reproduction

Type a single character in the `date-time-input`, such as '0', so that it would be filled like '01/01/2000' on losing focus.

### Workaround

Can be fixed by reassigning the `isSameValue` flag after the mask is parsed.

### Is this a regression?

### Affected versions

any

any

ddaribo:
Closes #1695

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Fixed
+- #### DateTime Input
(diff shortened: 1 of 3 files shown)


rkaraivanov (changes requested):
LGTM, however add a Changelog entry

rkaraivanov (approved):

PR #1701, opened 2025-05-22
fix(Calendar): Add tag for blazor gen as indirect render.

MayaKirova:
Related to [this](https://infragistics.visualstudio.com/NetAdvantage/_git/eded35d7-efa9-46e9-86b0-65a256568bb6/pullrequest/16605)

### DIFF
--- a/src/components/calendar/calendar.ts
+++ b/src/components/calendar/calendar.ts
@@ -17,6 +17,7 @@ import {
   pageUpKey,
   shiftKey,
 } from '../common/controllers/key-bindings.js';
+import { blazorIndirectRender } from '../common/decorators/blazorIndirectRender.js';
 import { watch } from '../common/decorators/watch.js';
 import { registerComponent } from '../common/definitions/register.js';
 import { IgcCalendarResourceStringEN } from '../common/i18n/calendar.resources.js';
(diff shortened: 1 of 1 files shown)


src/components/calendar/calendar.ts
```diff
@@ -105,6 +106,7 @@ export const focusActiveDate = Symbol();
  * @csspart selected - Indicates selected state. Applies to date, month and year elements of the calendar.
  * @csspart current - Indicates current state. Applies to date, month and year elements of the calendar.
  */
+@blazorIndirectRender
```
damyanpetev, line 109: This sould also work with the comment tag, no? Like `/* blazorIndirectRender */`?

damyanpetev: It shouldn't :) There's a diff with `/* not a doc comment */`  and `/** doc comment */` + those can be mixed like we do in Angular (see [action strip](https://github.com/IgniteUI/igniteui-angular/blob/88d98ab0e0b0fed08f27f1d60d38b6da1fdf4140/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts#L45) for example).

damyanpetev (approved):

PR #1704, opened 2025-05-23
Analyzer adjustments for 6.0

damyanpetev:


src/components/tooltip/tooltip.ts
```diff
@@ -21,9 +21,13 @@ import { all } from './themes/themes.js';
 import { styles } from './themes/tooltip.base.css.js';
 
 export interface IgcTooltipComponentEventMap {
+  /* blazorAlternateType: CustomEvent<void> */
```
damyanpetev, line 24: @rkaraivanov These don't actually do the thing it says, merely there as a reminder these are hard-remapped in the analyzer. But also... would consider actually changing the event details, because I'm not sure that details is useful.

rkaraivanov (approved):

PR #1747, opened 2025-06-14
fix(file-input): make all elements styleable with CSS variables + fix Bootstrap invalid box-shadow

### LINKED ISSUE #1676: [Bug]: Input group of type file has wrong validation styles in bootstrap theme

Input

### Description

The input has wrong validation styles; the shadow outline in valid/invalid  states should be shown only on focus

### Reproduction

Enable input validation and make the input invalid/valid 

### Workaround

No

### Is this a regression?

### Affected versions

All

Browser: All
OS: all

adrianptrv:
Closes #1676 

This PR is related to the fix: [#439](https://github.com/IgniteUI/igniteui-theming/pull/439)

I have added CSS variables for parts of the file-input that couldn't be styled individually before, using only variables.
I also fixed the box-shadow of the Bootstrap invalid state on the file-input, input, select, and textarea components so that it only appears when the component is focused, and not all the time.

### DIFF
--- a/src/components/file-input/themes/light/file-input.fluent.scss
+++ b/src/components/file-input/themes/light/file-input.fluent.scss
@@ -6,9 +6,3 @@ $theme: $fluent;
 :host {
     @include css-vars-from-theme(diff($base, $theme), 'ig');
 }
-
-:host(:is(igc-file-input)) {
-  [part~='file-names'] {
-      color: var(--ig-file-names-foreground, #000)
-  }
-}
--- a/src/components/file-input/themes/dark/file-input.fluent.scss
+++ b/src/components/file-input/themes/dark/file-input.fluent.scss
@@ -7,9 +7,3 @@ $theme: $fluent;
 :host {
     @include css-vars-from-theme(diff(light.$base, $theme), 'ig');
 }
-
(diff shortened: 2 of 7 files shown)


src/components/file-input/themes/light/file-input.fluent.scss
```diff
@@ -0,0 +1,14 @@
...
+
+$theme: $fluent;
+
+:host {
+    @include css-vars-from-theme(diff($base, $theme), 'ig');
+}
+
+:host(:is(igc-file-input)) {
+  [part~='file-names'] {
+      color: var(--ig-file-names-foreground, #000)
```
simeonoff: Shouldn't this come from the $theme? I see this repeated in all files and it doesn't sit right.

simeonoff:
> When I change the filename background, the input looks like this.
> 
> Screenshot 2025-07-02 at 11 23 00
> Is this the expected behavior?
> 
> To change the entire background of the file input, I need to modify the box-background variable, which is part of the input-group, to match the file-name-background, which is part of the file type.

You are changing the file names background so it looks correct to me.

simeonoff (approved):

PR #1774, opened 2025-07-03
fix(carousel): pause auto-rotation on pointer focus

### LINKED ISSUE #1731: Carousel resume animation behavior with focus
Okay, I get that's the current behavior replicating what the focus ring did, but it doesn't make sense. 
See https://igniteui.github.io/igniteui-webcomponents/?path=/story/carousel--inputs-template&args=interval:3000
Where focusing an input and moving the pointer away starts changing slides, which I don't think is nice and we should consider updating that behavior.

_Originally posted by @damyanpetev in https://github.com/IgniteUI/igniteui-webcomponents/pull/1684#discussion_r2129080103_

RivaIvanova:
Closes #1731

### DIFF
--- a/src/components/carousel/carousel.ts
+++ b/src/components/carousel/carousel.ts
@@ -105,11 +105,11 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
   private static readonly increment = createCounter();
   private readonly _carouselId = `igc-carousel-${IgcCarouselComponent.increment()}`;
 
+  private _paused = false;
   private _lastInterval!: ReturnType<typeof setInterval> | null;
   private _hasKeyboardInteractionOnIndicators = false;
-  private _hasMouseStop = false;
-  private _hasKeyboardFocus = false;
-  private _hasMouseFocus = false;
+  private _hasPointerInteraction = false;
+  private _hasInnerFocus = false;
 
   private _context = new ContextProvider(this, {
     context: carouselContext,
@@ -151,9 +151,6 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
 
   @state()
   private _playing = false;
-
-  @state()
-  private _paused = false;
 
   private _observerCallback({
     changes: { added, attributes },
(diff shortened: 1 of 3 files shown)


src/components/carousel/carousel.ts
```diff
@@ -329,14 +327,13 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
 
     addThemingController(this, all);
 
-    addSafeEventListener(this, 'pointerenter', this.handlePointerEnter);
-    addSafeEventListener(this, 'pointerleave', this.handlePointerLeave);
+    addSafeEventListener(this, 'pointerenter', this.handlePointerInteraction);
+    addSafeEventListener(this, 'pointerleave', this.handlePointerInteraction);
     addSafeEventListener(this, 'pointerdown', () => {
       this._hasInnerFocus = false;
     });
```
rkaraivanov: This can be removed since pointerdown falls into the focusin handler logic.

src/components/carousel/carousel.ts
```diff
@@ -389,42 +386,27 @@ export default class IgcCarouselComponent extends EventEmitterMixin<
...
 
-    if (this._hasInnerFocus) {
-      this._hasInnerFocus = false;
+    this._hasInnerFocus = event.type === 'focusin';
 
-      if (!this._hasMouseStop) {
-        this.handlePauseOnInteraction();
-      }
+    if (!this._hasPointerInteraction) {
+      this.handlePauseOnInteraction();
```
rkaraivanov, line 406: Maybe something along the lines of:

```ts
private _handleFocusState(event: FocusEvent): void {
    const isFocusin = event.type === 'focusin';
    const hasFocusWithin = this.contains(document.activeElement);

    if (isFocusin && this._hasInnerFocus) {
      return;
    }

    this._hasInnerFocus = hasFocusWithin;
    if (!(this._hasPointerInteraction && this._hasInnerFocus)) {
      this.handlePauseOnInteraction();
    }
  }
```

rkaraivanov (changes requested):

PR #1811, opened 2025-08-04
Thristodorova/chat expose default templates

teodosiah:


src/components/chat/chat-message.ts
```diff
@@ -79,33 +67,30 @@ export default class IgcChatMessageComponent extends LitElement {
...
+    return this.message
+      ? html`
+          <div part=${containerPart}>
+            <div part="bubble">
+              ${this._chatState?.options?.templates?.messageTemplate
+                ? this._chatState.options.templates.messageTemplate(
+                    this.message
+                  )
+                : this._chatState?.defaultMessageTemplate(this.message)}
+              ${this.message?.attachments &&
```
gedinakova: As we check for message initially (line 71), we don't need these further checks
```suggestion
              ${this.message.attachments &&
```

src/components/chat/chat-message.ts
```diff
@@ -79,33 +67,30 @@ export default class IgcChatMessageComponent extends LitElement {
...
+              this.message?.attachments.length > 0
```
gedinakova:
```suggestion
              this.message.attachments.length > 0
```

PR #1820, opened 2025-08-18
fix(input-group): bootstrap prefix & suffix border

### LINKED ISSUE #1817: [Bug]: Input Group - Styling issue in bootstrap theme

Combo

### Description

If you stack more than one prefix/suffix in the input group components, there is no border between them. 

### Reproduction

Add two prefixes and a suffix in the combo component and switch to the Bootstrap theme.

### Workaround

There should be a border between each prefix/suffix.

### Is this a regression?

### Affected versions

all

Image

adrianptrv:
Closes #1817

### DIFF
--- a/src/components/input/themes/shared/input.bootstrap.scss
+++ b/src/components/input/themes/shared/input.bootstrap.scss
@@ -10,7 +10,7 @@ $theme: $bootstrap;
     --input-font: #{sizable(rem(16px), rem(16px), rem(20px))};
     --input-border-color: #{var-get($theme, 'border-color')};
     --input-height: #{var-get($theme, 'size')};
-		--prefix-suffix-padding: #{pad-inline(rem(8px), rem(12px), rem(14px))};
+    --affix-padding: #{pad-inline(rem(8px), rem(12px), rem(14px))};
 
     ::part(helper-text) {
         @include type-style('body-2');
--- a/src/components/textarea/themes/shared/textarea.bootstrap.scss
+++ b/src/components/textarea/themes/shared/textarea.bootstrap.scss
@@ -7,7 +7,7 @@ $theme: $bootstrap;
     @include type-style('body-1');
 
     --component-size: var(--ig-size, #{var-get($theme, 'default-size')});
-    --prefix-suffix-padding: #{pad-inline(rem(10px), rem(12px), rem(12px))};
+    --affix-padding: #{pad-inline(rem(10px), rem(12px), rem(12px))};
 
     ::part(helper-text) {
         @include type-style('body-2');
@@ -15,11 +15,6 @@ $theme: $bootstrap;
         color: var-get($theme, 'helper-text-color');
         grid-auto-rows: minmax(rem(20px), auto);
         margin-block-start: rem(4px);
-    }
-
-    [part~='prefix'],
-    [part~='suffix'] {
-        padding-inline: unset;
     }
 
     ::slotted([slot="prefix"]),
--- a/src/components/combo/themes/shared/combo.common.scss
+++ b/src/components/combo/themes/shared/combo.common.scss
@@ -97,5 +97,5 @@ $dropdown-theme: dropdown-theme.$material;
 
 ::slotted([slot='suffix']),
 ::slotted([slot='prefix']) {
-    padding-inline: var(--prefix-suffix-padding);
+    padding-inline: var(--affix-padding);
 }
--- a/src/components/date-range-picker/themes/shared/date-range-picker.common.scss
+++ b/src/components/date-range-picker/themes/shared/date-range-picker.common.scss
@@ -132,5 +132,5 @@ igc-icon,
 
 ::slotted([slot|='suffix']),
 ::slotted([slot|='prefix']) {
-    padding-inline: var(--prefix-suffix-padding);
+    padding-inline: var(--affix-padding);
 }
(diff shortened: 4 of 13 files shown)


src/components/combo/themes/shared/combo.bootstrap.scss
```diff
@@ -17,6 +17,19 @@ $dropdown-theme: dropdown-theme.$bootstrap;
         grid-auto-rows: minmax(rem(20px), auto);
         margin-block-start: rem(4px);
     }
+
+    [slot='suffix']:not([part|='toggle'], [part|='clear']),
+    [slot='prefix'] {
+        padding-inline: unset;
```
simeonoff: I would use initial here, instead.

src/components/combo/themes/shared/combo.bootstrap.scss
```diff
@@ -17,6 +17,19 @@ $dropdown-theme: dropdown-theme.$bootstrap;
...
+    }
+
+    ::slotted([slot='suffix']),
+    ::slotted([slot='prefix']) {
+        padding-inline: pad-inline(rem(8px), rem(12px), rem(14px));
```
simeonoff: I don't like that we now have to type these values in several separate places (input, select, combo) and keep them in sync. A better approach would be to create an intermediary CSS custom property in the context of the igc-input component that can also be used here.

src/components/combo/themes/shared/combo.bootstrap.scss
```diff
@@ -128,4 +141,25 @@ $dropdown-theme: dropdown-theme.$bootstrap;
             color: inherit;
         }
     }
+
+    ::slotted([slot='suffix']),
+    [part|='toggle'] {
+        border-inline-end: .0625rem solid var(--disabled-border-color);
```
simeonoff: Please, use the rem function directly to represent values. Seeing rem(1px) is much easier to understand than .0625rem. Also, the benefit of using the rem function that we have is that the value is bound to the value of the global base font size. If you have issues with using the function just interpolate the call `#{rem(1px)} solid var(--disabled-border-color)`

src/components/combo/themes/shared/combo.bootstrap.scss
```diff
@@ -128,4 +141,25 @@ $dropdown-theme: dropdown-theme.$bootstrap;
...
+    }
+
+    ::slotted([slot='prefix']) {
+        border-inline-start: .0625rem solid var(--disabled-border-color);
```
simeonoff: Same as the comment above.

src/components/combo/themes/shared/combo.bootstrap.scss
```diff
@@ -128,4 +141,25 @@ $dropdown-theme: dropdown-theme.$bootstrap;
...
+:host(:not([disabled])),
+:host(:not(:disabled)) {
+    ::slotted([slot='suffix']),
+    [part|='toggle'] {
+        border-inline-end: .0625rem solid var(--border-color);
+    }
+
+    ::slotted([slot='prefix']) {
+        border-inline-start: .0625rem solid var(--border-color);
+    }
```
simeonoff: I guess you're trying to fix a regression with this code here? I.e. it's not related to the actual bug this PR addresses?

src/components/select/themes/shared/select.bootstrap.scss
```diff
@@ -17,6 +17,19 @@ $input-theme: input-theme.$bootstrap;
...
     }
+
+    [slot='suffix']:not([part|='toggle']),
+    [slot='prefix'] {
+        padding-inline: unset;
+    }
+
+    ::slotted([slot='suffix']),
+    ::slotted([slot='prefix']) {
+        padding-inline: pad-inline(rem(8px), rem(12px), rem(14px));
```
simeonoff: Same comment as with the combo.

src/components/combo/themes/shared/combo.bootstrap.scss
adrianptrv: It's part of the current fix, it applies the border at the end of every suffix element and at the start of every prefix element, respectively. Here's how it looks with and without the code:
img1

src/components/input/themes/shared/input.bootstrap.scss
```diff
@@ -10,6 +10,7 @@ $theme: $bootstrap;
     --input-font: #{sizable(rem(16px), rem(16px), rem(20px))};
     --input-border-color: #{var-get($theme, 'border-color')};
     --input-height: #{var-get($theme, 'size')};
+		--prefix-suffix-padding: #{pad-inline(rem(8px), rem(12px), rem(14px))};
```
simeonoff:
```suggestion
	--prefix-suffix-padding: #{pad-inline(rem(8px), rem(12px), rem(14px))};
```

src/components/textarea/themes/shared/textarea.bootstrap.scss
```diff
@@ -15,6 +16,18 @@ $theme: $bootstrap;
         grid-auto-rows: minmax(rem(20px), auto);
         margin-block-start: rem(4px);
     }
+
+    [part~='prefix'],
+    [part~='suffix'] {
+        padding-inline: unset;
+    }
```
simeonoff: Do you really need this here anymore?

simeonoff (approved):

PR #1835, opened 2025-08-29
feat(localization): Integration of new i18nManager.

skrustev:


rkaraivanov:
Abstract this into a controller class so each component can composite it into itself correctly.
There are also several other components such as combo and carousel which can use this functionality

src/components/common/i18n/i18n-controller.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** @internal */
+  public hostDisconnected(): void {
+    getI18nManager().removeEventListener(
+      'onResourceChange',
+      this._resourceChangeHandler
+    );
+  }
+
+  protected onResourceChange(event: CustomEvent<IResourceChangeEventArgs>) {
+    this._defaultResourceStrings = this.getCurrentResourceStrings();
```
MayaKirova: Shouldn't we set the new locale here? Currently when new `lang` is set on the html element it triggers the logic in core and raises the event, however `this.locale` remains the old one in the controller and the resource strings don't change.

Same happens if locale is set via the public API:

```
setCurrentI18n('bg');
```

Events fire but resources don't change.

skrustev: The locale for the component is resolved on line 33, because it can be overridden. It seems to be working on my end, do you have steps for reproducing?
Also note that for the calendar for example this PR is not changing the Intl implementations.
Also the stories override the locale and set it as input.

MayaKirova: Could just be the calendar, but if you run the calendar sample and change the closest html `lang` tag, most strings don't change:

image

Also if I set:

```
import {
  setCurrentI18n
} from 'igniteui-i18n-core';
setCurrentI18n('bg');
```
Inside stories\calendar.stories.ts and run the sample, the calendar remains in the default locale.

skrustev: I see, should fixed with this PR #1845
I haven't touched any of the locale handling in this PR like I said and the Calendar already had Intl used in its implementation, but the logic for setting locale was only local per component anyway so in the new PR it was replaced entirely and is obsolete.

PR #1845, opened 2025-09-10
Use new Intl formatter instances where applicable.

skrustev:


MayaKirova (dismissed):

src/components/date-time-input/date-time-input.ts
```diff
@@ -229,11 +229,10 @@ export default class IgcDateTimeInputComponent extends EventEmitterMixin<
   @watch('locale', { waitUntilFirstUpdate: true })
   protected setDefaultMask(): void {
     this.updateDefaultDisplayFormat();
-    this.updateDefaultMask();
-    this.setMask(this._defaultMask);
 
-    if (this.value) {
-      this.updateMask();
+    if (!this._inputFormat) {
```
skrustev, line 225: Tried to separate the commit but didn't work as expected... Old Lines 235-236 should remain and are added back in next commit.

src/components/combo/combo.ts
```diff
@@ -326,11 +327,14 @@ export default class IgcComboComponent<
    * The resource strings for localization.
    */
   @property({ attribute: false })
-  public set resourceStrings(value: IComboResourceStrings) {
+  public set resourceStrings(
+    value: IComboResourceStrings & IValidationResourceStrings
```
rkaraivanov: I thought we agreed to drop the `IValidationResourceStrings`

src/components/calendar/days-view/days-view.ts
```diff
@@ -336,7 +331,10 @@ export default class IgcDaysViewComponent extends EventEmitterMixin<
   }
 
   private intlFormatDay(day: CalendarDay) {
-    const fmt = this._intl.get('label');
+    const fmt = getDateFormatter().getIntlFormatter(
```
rkaraivanov, line 334: Nitpick:
This is no way related to the PR but from a DX point of view
won't it be better that `getDateFormatter` accepts the locale and the DateTimeFormatOptions and skips the chaining invocation of `getIntlFormatter`.

Something to ponder regarding the i18n core package public API.

src/components/combo/combo.ts
```diff
@@ -921,6 +927,7 @@ export default class IgcComboComponent<
         part="clear-icon"
         @click=${this.handleClearIconClick}
         ?hidden=${this.disableClear || this._selection.isEmpty}
+        aria-label=${this.resourceStrings.combo_clearItems_placeholder!}
```
rkaraivanov, line 930: Nitpick:
Maybe wrap it in `ifDefined` if that can truly be undefined

src/components/date-time-input/date-time-input.spec.ts
```diff
@@ -60,25 +60,27 @@ describe('Date Time Input component', () => {
...
+      expect(el.displayFormat).to.equal('d.M.yyyy');
+      expect(input.value).to.equal('3.3.2020');
     });
 
     it('should use inputFormat if no displayFormat is defined - issue 1114', async () => {
       el.inputFormat = 'yyyy#MM#dd';
       await elementUpdated(el);
 
-      expect(el.displayFormat).to.be.undefined;
+      expect((el as any)._displayFormat).to.be.undefined;
```
rkaraivanov, line 83: Why are we testing this internal API?

src/components/file-input/file-input.ts
```diff
@@ -1,9 +1,13 @@
+import {
+  type IInputResourceStrings,
+  InputResourceStringsEN,
```
rkaraivanov: Unfortunate naming as I would expect FileInputResourceStrings<Locale> but I guess that is because of legacy

src/components/calendar/days-view/days-view.ts
skrustev: That is because there's was no implemented formatRange in our Date Formatter. I could expose it since exposing of methods is on a need to have basis. Wanted to change existing logic as less as possible, but this shouldn't cause any difference.

src/components/combo/combo.ts
skrustev: It shouldn't be undefined in general, as it should always return at least default EN resources.

src/components/calendar/days-view/days-view.ts
skrustev: And it seems like the previous implementation reuses the Intl formatter without calling the format method, depending on the case. It is a bit more optimized that way and I can see how this can be a way to use the formatter as well.

src/components/date-time-input/date-time-input.spec.ts
skrustev: `displayFormat` returns the resolved display format despite if it defined by user or not. This check pretty ensures that for this test it is not set by the 'user' initially.

src/components/file-input/file-input.ts
skrustev: Dang, I'll rename it in the core package now since it appear previously was wrong anyway.

rkaraivanov (approved):

PR #1880, opened 2025-10-07
feat: Added highlight container

rkaraivanov:


src/components/highlight/service.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  private _createRegex(value: string): RegExp {
+    return new RegExp(
+      escapeRegex(value),
+      this._host.caseSensitive ? 'dg' : 'dgi'
+    );
+  }
+
+  private _getRangeByIndex(index: number): Range {
+    return this._highlight.values().drop(index).next().value as Range;
```
simeonoff: @rkaraivanov This iterates over the entire set each time when called in goTo... so basically in each `next`, `previous`, `setActive`. It's probably fine for smaller documents with a few matches, but for large documents it could quickly get out of hand IMO. Instead we could probably have an internal cache in the service (`private _ranges = Range[]`) that we can push to every time we add to `_highlight`? Then in `_getRangeByIndex` we could simply return `this._ranges[index]`. Finally, we reset the cache array in the `clear` method.

What do you think?

src/components/highlight/service.ts
```diff
@@ -0,0 +1,245 @@
...
+   * Finds matches for the given search text in the component's content and creates highlight ranges for them.
+   */
+  public find(value: string): void {
+    if (!value) {
+      return;
+    }
+
+    const nodes = iterNodes(this._host, {
+      show: 'SHOW_TEXT',
+      filter: (node) => !!node.textContent,
```
simeonoff: You already check if the node has textContent in the matchText util for each node. Don't know how useful this filter operation is.

PR #1885, opened 2025-10-13
fix(date-picker): Emit change event in non-editable configurations

### LINKED ISSUE #1884: [Bug]: IgcDatePicker - igcChange event is not triggered when the clear button is clicked

Date picker

### Description

For IgcDatePicker with mode="dialog", igcChange event is not triggered when the clear button is clicked.

### Reproduction

1. Go to the sample https://codesandbox.io/p/devbox/damp-sun-q624vl?workspaceId=ws_Wca1rSR5QWnYWuPzcJ6uNq
2. Click the clear button on IgbDatePicker.
3. Observe the browser console.

Actual Result:
No output on the console.

Expected Result:
"igcChange" appears on the console.

### Workaround

I have worked around this problem by templating a clear button.

### Is this a regression?

### Affected versions

Failing in 6.3.2

Chrome: 141.0.7390.65

rkaraivanov:
Date-picker correctly emits `igcChange` event when in dialog mode or when set to a non-editable configuration, such as readonly or non-editable, after user interaction.

Closes #1884

### DIFF
--- a/src/components/date-picker/date-picker.ts
+++ b/src/components/date-picker/date-picker.ts
@@ -600,7 +600,6 @@ export default class IgcDatePickerComponent extends FormAssociatedRequiredMixin(
 
   protected _handleDialogClosed(event: Event): void {
     event.stopPropagation();
-    this._oldValue = this.value;
   }
 
   //#endregion


src/components/date-picker/date-picker.ts
```diff
@@ -593,11 +594,13 @@ export default class IgcDatePickerComponent extends FormAssociatedRequiredMixin(
...
   protected _handleDialogClosing(event: Event): void {
     event.stopPropagation();
+    this._oldValue = this.value;
     this._hide(true);
   }
 
   protected _handleDialogClosed(event: Event): void {
     event.stopPropagation();
+    this._oldValue = this.value;
   }
```
ddaribo, line 603: Probably redundant to assign `_oldValue` in both close events?
Otherwise the fix & refactoring LGTM. 
I noticed a small issue with the label positioning after value is cleared (DP&DRP, material, outlined), but I'll log it separately since it is not related to the current PR.

```suggestion
  protected _handleDialogClosing(event: Event): void {
    event.stopPropagation();
    this._oldValue = this.value;
    this._hide(true);
  }

  protected _handleDialogClosed(event: Event): void {
    event.stopPropagation();
  }
```

PR #1894, opened 2025-10-15
fix(datepicker): fix datepicker issues when clearing value and notch border

### LINKED ISSUE #1886: [Bug]: IgcDatePicker, IgcDateRangePicker - material, outlined label positioning issue after value is cleared

Date picker, Date range picker

### Description

Similar to [1791](https://github.com/IgniteUI/igniteui-webcomponents/issues/1791)

For material theme, `outlined=true`, the labels of the date pickers are misplaced in two different ways after value is cleared, depending on whether the  input is editable or not. 

### Reproduction

DatePicker - `nonEditable=true` & `mode='dropdown'` OR  `nonEditable=true|false` & `mode='dialog'`:

Image

DatePicker - `nonEditable=false` & `mode='dropdown'`:

Image

Same for DateRangePicker when `useTwoInputs=true`.

### Workaround

### Is this a regression?

### Affected versions

latest

Browser: Chrome Version 141.0.7390.66 (Official Build) (64-bit)

didimmova:
Closes #1886


src/components/date-picker/date-picker.ts
```diff
@@ -611,6 +611,11 @@ export default class IgcDatePickerComponent extends FormAssociatedRequiredMixin(
     this._oldValue = this.value;
     this.value = null;
     this._input?.clear();
+
+    if (this.nonEditable || !this._isDropDown) {
+      this._input.blur();
+      Promise.resolve().then(() => this._input.focus());
+    }
```
ddaribo: This seems to fix the positioning issue on clear through the icon, however, I am not convinced it would be the best solution.

In input.material.scss lines 145-159, there are the following styles, which set the proper positioning when the input is not `readonly`. Having in mind that the issue occurs for `readonly`, I believe removing the :not([readonly]) selector might address it?

```scss
:host(:not([readonly]):focus-within) {
    [part='label'] {
        @include type-style('caption');

        translate: 0 -73%;
    }
}

:host([outlined]:not([readonly]):focus-within) {
    [part='label'] {
        @extend %label;

        align-self: start;
    }
}
```

src/components/date-range-picker/themes/shared/date-range-picker.common.scss
```diff
@@ -60,6 +60,12 @@ $theme: $material;
     }
 }
 
+:host([outlined]) {
+    igc-date-time-input::part(notch) {
+        border-top: 2px solid transparent;
```
simeonoff:
```suggestion
        border-top: rem(2px) solid transparent;
```

src/components/date-picker/themes/shared/date-picker.material.scss
```diff
@@ -3,6 +3,12 @@
 
 $theme: $material;
 
+:host([outlined]) {
+    igc-date-time-input::part(notch) {
+        border-top: 2px solid transparent;
```
simeonoff:
```suggestion
        border-top: rem(2px) solid transparent;
```

PR #1897, opened 2025-10-20
feat(checkbox): set tick-width property

adrianptrv:
Sets the newly added tick-width property to be used by the component.

Depends on: [#485](https://github.com/IgniteUI/igniteui-theming/pull/485)


desig9stein:
@adrianptrv I am unable to change the --tick-width for the indigo theme since the prop is overridden.

Screenshot 2025-10-22 at 16 42 25

simeonoff (dismissed):

Issue #1915, opened 2025-10-27
[Bug]: Tabs not fully styleable using Sass theme


simeonoff:
WC: $item-icon-color, $item-hover-icon-color, $item-active-icon-color, $item-active-background, $indicator-color not working

PR #1929, opened 2025-10-29
fix(tabs): fix non-working Sass theme parameters

adrianptrv:
Closes: #1915 
**Requires this theming PR: [495](https://github.com/IgniteUI/igniteui-theming/pull/495)**

- Replaced `igx-icon` with `igc-icon` in the Tabs WC demo (Angular project) to ensure the `Tabs` theme "icon" parameters work correctly. [#16395](https://github.com/IgniteUI/igniteui-angular/pull/16395)
- Also set up the newly added theme parameters - `$item-active-hover-background` `$item-active-hover-color` `$item-active-hover-icon-color`


adrianptrv:
> THEME: INDIGO It looks like `item-active-background` is not used anywhere in the Indigo theme, including in the Angular repo. Its current value also seems incorrect: when I apply it in the expected place, it renders as transparent. I believe it should use the same color as the `item-hover-background`, since that’s what is currently used in that state.

I have added and set up three new parameters for the active tab - `$item-active-hover-background` `$item-active-hover-color` `$item-active-hover-icon-color` 
The changes can now be tested again using this theming PR: [495](https://github.com/IgniteUI/igniteui-theming/pull/495)

PR #1969, opened 2025-11-13
Splitter component

MonikaKirkova:


src/components/common/context.ts
rkaraivanov, line 1: Since you're not longer using Lit context, the changes here should be reverted.

src/components/resize-container/resize-controller.ts
rkaraivanov, line 1: Since you're not using the resize controller, the changes here should be reverted.

src/components/resize-container/types.ts
rkaraivanov, line 1: Same as the resize controller changes

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,955 @@
...
+  private _resizeState: SplitterResizeState | null = null;
+  private _isDragging = false;
+  private _dragPointerId = -1;
+  private _dragStartPosition = { x: 0, y: 0 };
+
+  private readonly _internals = addInternalsController(this, {
+    initialARIA: {
+      ariaOrientation: 'horizontal',
+    },
+  });
```
rkaraivanov: You don't need the internals controller; all relevant ARIA attributes should be applied directly on the splitter bar, since that is the only focusable part of the element by default.

rkaraivanov (changes requested):

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1031 @@
...
+          this._handleExpanderClick('start', e)}
+      ></div>
+      <div part="drag-handle" ?hidden=${dragHandleHidden}></div>
+      <div
+        part="${partMap(this._resolvePartNames('end'))}"
+        ?hidden=${nextButtonHidden}
+        role="img"
+        aria-label=${this._resolveExpanderLabel('end')}
+        @pointerdown=${(e: PointerEvent) => this._handleExpanderClick('end', e)}
+      ></div>
```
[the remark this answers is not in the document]
ddaribo: leaving the button with `role="image"`, as the keyboard interactivity is done through the bar

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1031 @@
...
+        @pointerdown=${(e: PointerEvent) =>
+          this._handleExpanderClick('start', e)}
+      ></div>
+      <div part="drag-handle" ?hidden=${dragHandleHidden}></div>
+      <div
+        part="${partMap(this._resolvePartNames('end'))}"
+        ?hidden=${nextButtonHidden}
+        role="img"
+        aria-label=${this._resolveExpanderLabel('end')}
+        @pointerdown=${(e: PointerEvent) => this._handleExpanderClick('end', e)}
```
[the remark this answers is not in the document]
ddaribo: leaving the button with `role="image"`, as the keyboard interactivity is done through the bar

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1045 @@
...
+    ) {
+      this._resetPanes();
+    }
+
+    if (this.hasUpdated) {
+      this._updatePanes();
+    }
+    super.update(changed);
+  }
+
```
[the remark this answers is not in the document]
ddaribo: Leaving this as is for now, as the initial state seems properly applied and no additional scheduled updates warnings are thrown with the current approach.

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1030 @@
...
+  endPanelSize: number;
+  /** The change in size since the resize operation started (only for igcResizing and igcResizeEnd) */
+  delta?: number;
+}
+
+export interface IgcSplitterComponentEventMap {
+  igcResizeStart: CustomEvent<IgcSplitterResizeEventDetail>;
+  igcResizing: CustomEvent<IgcSplitterResizeEventDetail>;
+  igcResizeEnd: CustomEvent<IgcSplitterResizeEventDetail>;
+}
```
rkaraivanov, line 48: This type should be exported in the relevant section of the `index.ts`.

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1030 @@
...
+import { all } from './themes/themes.js';
+
+export interface IgcSplitterResizeEventDetail {
+  /** The current size of the start panel in pixels */
+  startPanelSize: number;
+  /** The current size of the end panel in pixels */
+  endPanelSize: number;
+  /** The change in size since the resize operation started (only for igcResizing and igcResizeEnd) */
+  delta?: number;
+}
```
rkaraivanov, line 42: This one as well.

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1030 @@
...
+        aria-label=${this._resolveExpanderLabel('start')}
+        @pointerdown=${(e: PointerEvent) =>
+          this._handleExpanderClick('start', e)}
+      ></div>
+      <div part="drag-handle" ?hidden=${dragHandleHidden}></div>
+      <div
+        part="${partMap(this._resolvePartNames('end'))}"
+        ?hidden=${nextButtonHidden}
+        role="img"
+        aria-label=${this._resolveExpanderLabel('end')}
```
rkaraivanov: Role `img` is usually discouraged since `/<picture>` has it implicitly. Usually what you would use would be `presentation` which has the same semantics 
as `img`. NVDA (guess most screen readers also) skip over this entirely so you can drop the role and aria-label attributes.

src/components/splitter/splitter.ts
```diff
@@ -0,0 +1,1030 @@
...
+    return value ? this._sizeToPercent(value) : defaultValue;
+  }
+
+  private _updateBarAria(): void {
+    const bar = this._barRef.value;
+    if (!bar) return;
+
+    bar.setAttribute('aria-valuenow', String(this._getStartPaneSizePercent()));
+    bar.setAttribute('aria-valuemin', String(this._getMinMaxAsPercent('min')));
+    bar.setAttribute('aria-valuemax', String(this._getMinMaxAsPercent('max')));
```
rkaraivanov:
```suggestion
    bar.ariaValueNow = this._getStartPaneSizePercent().toString();
    bar.ariaValueMin = this._getMinMaxAsPercent('min').toString();
    bar.ariaValueMax = this._getMinMaxAsPercent('max').toString();
```

rkaraivanov (changes requested):
Good job with the changes.

I have some last "requests" 😄 which I've left as comments and these one 
below:

I'd like you to move all `@property` decorators on the **setter** of a given accessor pair as per the [official docs](https://lit.dev/docs/components/properties/#accessors:~:text=To%20use%20custom%20property%20accessors%20with%20the%20%40property%20or%20%40state%20decorators%2C%20put%20the%20decorator%20on%20the%20setter%2C%20as%20shown%20above%2E%20%40property%20or%20%40state%20decorated%20setters%20call%20requestUpdate%28%29%2E).

Also, you can drop the **reflect** option on the non-boolean properties.
I don't think it brings much value aside from having to write back to the DOM of the component.

PR #2020, opened 2025-12-05
feat(i18n): Update packages to release version.

skrustev:


rkaraivanov (changes requested):
LGTM

Just add the core localization package to the [**real**](https://github.com/IgniteUI/igniteui-webcomponents/blob/master/scripts/_package.json) package.json as well.

I suppose you can also add the resources package as an optional peer dependency in, just so users can grok what they need to install if they want to use our predefined resources.

PR #2028, opened 2025-12-09
fix(tabs): fix selected indicator alignment

### LINKED ISSUE #2027: [Bug]: Tabs – When a `text-align` CSS property is applied to a parent element, the Tabs indicator also aligns according to it.

Tabs & Tab group

### Description

When a `text-align` CSS property is applied to a parent element, the Tabs indicator also inherits it and becomes misaligned with the selected tab. The issue occurs only in Web Components and affects all themes except the Bootstrap one.

https://github.com/user-attachments/assets/916f71f1-1f23-4c90-a378-0e2e8ba9d863

### Reproduction

1. Open a Tabs sample.
2. Select the parent element of the `igc-tabs` component.
3. Apply a `text-align` property to the parent element.

### Workaround

### Is this a regression?

### Affected versions

Failing in 6.0.0 and onwards. Worked in 5.4.0.

Browser: Chrome - Version 142.0.7444.176 (Official Build) (arm64)
OS: macOS 15.6

didimmova:
Closes #2027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,80 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   - New localization implementation. Now you can set global localization for all localized Ignite UI components using the `registerI18n` and `setCurrentI18n` methods.
   - Calendar, Date Input, Date Picker and Date Range Picker components now implement the new localization for formatting as well, which internally uses new implementation of the Intl.
   - Added localization for Carousel, Chip, Combo, File Input and Tree. They now have `locale` and `resourceStrings` properties as well, which you can set to customize them individually.
+- #### Badge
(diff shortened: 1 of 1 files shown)


src/components/tabs/themes/tabs.base.scss
```diff
@@ -136,6 +136,7 @@ igc-icon-button::part(base) {
     position: relative;
     grid-row: 2;
     grid-column: 1 / -1;
+    text-align: initial;
```
desig9stein, line 140: I saw that the body content also gets centered, so I think it's better to have `text-align: initial` on the host(igx-tabs)

desig9stein: Or if you can't add it on the host, you can also add it on the body, but you have to do it for the body in Angular as well

PR #2033, opened 2025-12-11
fix(carousel): improve carousel context handling with async context

IMinchev64:
Closes [this Bug](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/41137)

Current carousel context wiring:
- The carousel sets up a Lit `ContextProvider` (`carouselContext`) when it constructs. Slides use `@consume({ context: carouselContext, subscribe: true })` to grab a `_carousel` ref.
- The consumer request fires once when the slide connects. If the carousel isn't upgraded/publishing yet (common if Blazor renders wrappers and upgrades happen later), that request is missed and `_carousel` stays undefined.

Why this causes issues (Blazor timing):
- If I understand correctly, Blazor renders web components, then the custom elements upgrade. Slides can connect before the carousel has attached its provider. Because the consume decorator issues a one-time request, there's no retry, so the ref never gets set. Downstream slide logic that expects `_carousel` blows up.

Why refactor following tile manager:
- Tile manager uses an async context consumer created after `updateComplete`, and the provider re-sets its value once ready. This avoids the race.
- Copying that pattern (or otherwise deferring/retrying the context request) ensures slides always get a valid carousel ref, eliminating the timing gap seen in Blazor.


src/components/carousel/carousel-slide.ts
```diff
@@ -38,8 +38,16 @@ export default class IgcCarouselSlideComponent extends LitElement {
...
+  private _carousel?: IgcCarouselComponent;
+
+  // Set carousel reference once provider is ready
+  private readonly _context = createAsyncContext(
+    this,
+    carouselContext,
+    (carousel) => {
+      this._carousel = carousel;
+    }
+  );
```
rkaraivanov: You can instantiate this inside the constructor of the slide element without binding it to an instance variable if you are not using it.

PR #2054, opened 2026-01-12
feat: Simplified Markdown renderer setup

### LINKED ISSUE #2053: Provide some pre-baked utilities for chat extras for external use
Related to https://github.com/IgniteUI/igniteui-angular/issues/16717 and the consecutive fix we settled on for now https://github.com/IgniteUI/igniteui-angular/pull/16729

It'd be much cleaner if the extras entry had exports for some pre-made basic marked/shiki setup that can directly be reused externally, so other packages don't need to specify peer dependencies for those packages as well, since that either restricts what `igniteui-webcomponents` can specify as peer without causing issues down the chain or what version of it other packages can rely on - neither of which seems flexible enough. Might be something else I'm missing, just noting this down for consideration as some form of enhancement.

rkaraivanov:
Closes #2053

### DIFF
--- a/src/extras/chat-markdown-renderer.ts
+++ b/src/extras/chat-markdown-renderer.ts
@@ -12,38 +12,10 @@ const DEFAULT_THEME = {
 };
 
 /**
- * Options to configure the MarkdownMessageRenderer.
- */
-export interface MarkdownRendererOptions {
-  /**
-   * If true, disables syntax highlighting entirely.
-   */
-  noHighlighter?: boolean;
-
-  /**
-   * List of programming languages to support in syntax highlighting.
-   */
-  languages?: string[];
-
-  /**
-   * The theme used by the syntax highlighter (e.g., 'github-light').
-   */
-  theme?: {
-    light?: string;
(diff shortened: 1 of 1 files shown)


src/extras/chat-markdown-renderer.ts
```diff
@@ -104,9 +190,38 @@ export async function createMarkdownRenderer(
...
+ *
+ * @example
+ * ```typescript
+ * const renderer = await createMarkdownRenderer();
+ * const message = { text: '# Hello **World**' };
+ * const template = await renderer(message);
+ * ```
+ */
+export async function createMarkdownRenderer(
+  options?: MarkdownRendererOptions
```
damyanpetev, line 192: Just one bit I don't get - why keep both `MarkdownRendererOptions` and `MarkdownSetupOptions` that don't inherit each other and the only diff is the addition of `colorReplacements` option and looking at the intended React integration that plans on using the -Setup type, I don't see the need for both and detached. If we want a door for extra stuff on the rendered, it could at least inherit the setup type, no?

PR #2056, opened 2026-01-14
feat: Added theme provider context for scoped theming support

rkaraivanov:


simeonoff (dismissed):

src/components/theme-provider/theme-provider.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  private _getContextValue(): ThemeContext {
+    return {
+      theme: this.theme,
+      variant: this.variant,
+    };
+  }
+
+  protected override render(): TemplateResult {
+    return html`<slot></slot>`;
```
damyanpetev, line 117: Hm, what use do we loose if this drops to Light DOM instead? Not a huge fan of enforcing a shadow root, especially if this might end up being used on application root as means to specify the initial theme even. Perhaps optionally and control `createRenderRoot`?

rkaraivanov: What is the concern here, all the content is in the light DOM anyway, since it is slotted.

src/components/common/definitions/defineAllComponents.ts
```diff
@@ -140,6 +141,7 @@ const allComponents: IgniteComponent[] = [
   IgcTileComponent,
   IgcTileManagerComponent,
   IgcTooltipComponent,
+  IgcThemeProviderComponent,
```
damyanpetev: I'd put this first actually - specifically, due to order of custom elements upgrading when the DOM precedes the scripts - the components looking for the provider _can_ initialize before it and fail to hook into context. You can repro that in the story if you move the define later like:
```ts
export const Basic: Story = {
  args: {
    theme: 'material',
    variant: 'light',
  },
  render: Template,
  play: () => {
    defineComponents(
      IgcButtonComponent,
      IgcInputComponent,
      IgcCardComponent,
      IgcCardHeaderComponent,
      IgcCardContentComponent,
      IgcDatePickerComponent,
      IgcThemeProviderComponent,
    );
  }
};
```
That breaks the provider context for the inner components.
image

damyanpetev: Not sure if we can ensure the provider is first when users go though per-component defines like that too. Other than documentation note, I don't see many great options.

src/theming/context.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ *   render() {
+ *     const theme = this._themeConsumer.value;
+ *     // ...
+ *   }
+ * }
+ * ```
+ */
+export const themeContext = createContext<ThemeContext>(
+  Symbol('igc-theme-context')
```
damyanpetev: This makes this context effectively a singleton for the entire app relying on a single `igniteui-webcomponents` package to function correctly, right? Cool in general, except I have a mild concern of slight version mismatches that can cause multiple versions and thus separate context Symbols. Can we consider global ones w/ `Symbol.for()` or really just a string token? Not sure it's that critical to have that internalized.

Issue #2057, opened 2026-01-15
[Bug]: Slotted validation text does not conform to the theme colors


rkaraivanov:
### Which component(s) are affected?

Checkbox / Switch

### Description

Slotted text inside checkbox/radio(s) validation slots is not using the invalid color from the current theme.

### Reproduction

1. Open the checkbox form story/radio group story.
2. Try to submit the form
3. Observe the error messages styles.

### Workaround

I have not found a workaround.

### Is this a regression?

No or unsure. This never worker, or I haven't tried before.

### Affected versions

Failing in master

### Browser/OS/Node environment

Browser: Latest Chrome

PR #2063, opened 2026-01-20
fix(validator): fix validator theme deriving input theme values

didimmova:
Closes https://github.com/IgniteUI/igniteui-webcomponents/issues/2057


rkaraivanov:
@didimmova 
image

A question for the Bootstrap dark variant, are the colors of the label(s) and the invalid slot text supposed to differ?

didimmova:
> @didimmova image
> 
> A question for the Bootstrap dark variant, are the colors of the label(s) and the invalid slot text supposed to differ?

this is something the design team should decide @sdimchevski and will be fixed in the `igniteui-theming` repo later if necessary

PR #2099, opened 2026-02-16
feat(splitter): add styles for the splitter

### LINKED ISSUE #2022: Splitter Styles implementation
[Design Hand-off](https://www.figma.com/design/m26ZFquPMB9hDpR4qExBFm/Splitter-Handoff?node-id=32-1883&t=Zrigzz8en67qjp8v-0)

didimmova:
Test with a package from this theming PR - https://github.com/IgniteUI/igniteui-theming/pull/533

Closes #2022

### DIFF
--- a/src/components/splitter/themes/shared/splitter.common.scss
+++ b/src/components/splitter/themes/shared/splitter.common.scss
@@ -81,12 +81,6 @@ $expander-border-active: $bar-size solid var-get($theme, 'expander-color-active'
     }
 }
 
-:host([orientation='horizontal'][hide-collapse-buttons][hide-drag-handle]) {
-    [part='splitter-bar'] {
-        width: $bar-size;
-    }
-}
-
 :host([orientation='vertical']) {
     [part='splitter-bar'] {
         height: $bar-size;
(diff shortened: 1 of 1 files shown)


src/components/splitter/themes/shared/splitter.common.scss
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            [part='start-collapse-btn'],
+            [part='end-expand-btn'] {
+                border-block-end: $expander-border-active;
+            }
+        }
+    }
+
+    [part='drag-handle'] {
```
desig9stein, line 102: There is no need to duplicate this in both horizontal and vertical orientations; add it under the `:host([orientation])` so it works for both orientations

src/components/splitter/themes/shared/splitter.common.scss
```diff
@@ -0,0 +1,133 @@
...
+        border-inline-start: $expander-border;
+    }
+
+    [part='start-collapse-btn'],
+    [part='end-expand-btn'] {
+        border-inline-end: $expander-border;
+    }
+}
+
+:host([orientation='horizontal'][hide-collapse-buttons][hide-drag-handle]) {
```
desig9stein: There is no need to duplicate this in both horizontal and vertical orientations; add it under the `:host([orientation][hide-collapse-buttons][hide-drag-handle])` so it works for both orientations

src/components/splitter/themes/shared/splitter.common.scss
didimmova: in one case, it's setting the height, in the other, it's setting the width

PR #2225, opened 2026-05-13
feat: Invoker commands integration for components

rkaraivanov:
## Description

This commit integrates invoker commands into our components, allowing for more flexible and dynamic interactions.

The changes include updates to button components,
the addition of a new command controller,
and modifications to related stories and configurations.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (code improvements without functional changes)

## Checklist

- [x] I have updated documentation if needed
- [ ] Breaking changes are documented in the description


src/components/dialog/dialog.ts
```diff
@@ -80,92 +126,96 @@ export default class IgcDialogComponent extends EventEmitterMixin<
...
+      .set('--hide', this.hide)
+      .set('--toggle', this.toggle);
   }
 
-  protected override firstUpdated(): void {
-    if (this.open) {
-      this._dialog?.showModal();
+  protected override updated(properties: PropertyValues<this>): void {
+    if (properties.has('open')) {
+      this.open ? this._dialog?.showModal() : this._dialog?.close();
```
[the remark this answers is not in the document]
rkaraivanov: Unless someone uses the internal API of the component, there is not invariant where the internal dialog will be in a state not controlled by the `open` property.

src/components/banner/banner.ts
```diff
@@ -54,8 +69,15 @@ export default class IgcBannerComponent extends EventEmitterMixin<
...
   /**
-   * Determines whether the banner is being shown/hidden.
-   * @attr
+   * Whether the banner is open.
+   *
+   * Setting this property programmatically will immediately show or hide the
+   * banner without animation and without emitting `igcClosing` / `igcClosed`.
+   * Prefer the `show()`, `hide()`, and `toggle()` methods for animated
+   * transitions. Events are only emitted when the banner is closed through
+   * user interaction with the default action button.
```
damyanpetev:
```suggestion
   * Setting this property programmatically will immediately show or hide the
   * banner without animation and without emitting close events.
   * Prefer the `show()`, `hide()`, and `toggle()` methods for animated
   * transitions.
```
Or just events, since there are no other events on the banner? Also the when those emit is should be in their description, otherwise we should put such notice on the methods as well. And then go and do that for the rest of the comps. Pass. 

PS: Is that a remark - as in do we put it in a `@remarks` tag?

src/components/button/button-base.ts
```diff
@@ -84,35 +126,84 @@ export abstract class IgcButtonBaseComponent extends EventEmitterMixin<
...
+  }
+
+  public get commandfor(): string | null {
+    return this._commandfor;
+  }
+
+  /**
+   * The target element for the invoker command. Resolved from the `commandfor` ID.
+   */
+  public get commandForElement(): Element | null {
```
damyanpetev, line 163: Think you can cs-suppress those pretty much immediately for the analyzer

src/components/button/button-base.ts
```diff
@@ -121,41 +212,72 @@ export abstract class IgcButtonBaseComponent extends EventEmitterMixin<
...
+    }
+  }
+
+  //#region Public API
+
+  /* alternateName: focusComponent */
+  /** Moves keyboard focus to the button. Accepts an optional `FocusOptions` object (e.g. `{ preventScroll: true }`). */
+  public override focus(options?: FocusOptions): void {
+    this._nativeButton?.focus(options);
+  }
```
damyanpetev: Um there's already `delegatesFocus` doing that? What's the point of the override?
~Also if its a href anchor shouldn't this check and call on that instead?~
Nvm, just saw this is a move of old code and the `_nativeButton` is named _and_ typed wrong, since it's just the base part. Can disregard, though really wouldn't mind an internal comment why  this is added :)

src/components/button/icon-button.ts
```diff
@@ -17,10 +17,50 @@ import { styles as shared } from './themes/icon-button/shared/icon-button.common
 import { all } from './themes/icon-button/themes.js';
 
 /**
+ * A button that displays a single icon, designed for compact, icon-only
+ * interactions such as toolbar actions, floating action buttons, or inline
+ * controls.
+ *
+ * The icon is sourced from the icon registry via the `name` and `collection`
+ * attributes. Like `igc-button`, it can render as an anchor element when
```
damyanpetev:
```suggestion
 * attributes. Like the normal button, it can render as an anchor element when
```

src/components/button/icon-button.ts
```diff
@@ -17,10 +17,50 @@ import { styles as shared } from './themes/icon-button/shared/icon-button.common
...
+ * <igc-icon-button
+ *   name="download"
+ *   href="/files/report.pdf"
+ *   download
+ *   variant="flat"
+ * ></igc-icon-button>
+ *
+ * @example
+ * <!-- Disabled icon button -->
+ * <igc-icon-button name="delete" disabled></igc-icon-button>
```
damyanpetev:
```suggestion
```
for xplat-reasons I'd keep this clean for now

src/components/common/controllers/id-resolver.ts
```diff
@@ -0,0 +1,162 @@
...
+   * Increment the reference count. Starts the underlying MutationObserver on the first call.
+   */
+  public retain(): void {
+    if (this._refCount++ === 0) {
+      this._observer?.observe(document.body, {
+        attributeFilter: ['id'],
+        attributeOldValue: true,
+        subtree: true,
+        childList: true,
+      });
```
damyanpetev: Um shouldn't we observe under the `getRootNode()` of the element with `commandfor=` or another ID ref?
Unless I'm missing something, this will observe under documents light DOM, so if the idref is inside a shadow root, it won't see it. And yeah, that'll probably mean a singleton observer per document root :/

damyanpetev: PS: Also not sure if safe to use `document.body` without checks - pretty sure this can still hit and not sure lit's `isServer` would actually cover our use cases.

src/components/common/controllers/id-resolver.ts
```diff
@@ -0,0 +1,162 @@
...
+    if (this._connected) {
+      ID_REF_CHANGE_EMITTER.removeEventListener('id-refs-change', this);
+      ID_REF_CHANGE_EMITTER.release();
+    }
+  }
+
+  /** Resolve an ID string to an element, scoped to the host's root node. */
+  public resolve(id: string): Element | null {
+    return getElementByIdFromRoot(this._host, id);
+  }
```
damyanpetev: Nit: Not sure why this is even a thing on the controller, it doesn't really do anything with the controller state. `getElementByIdFromRoot` is already a static util elements can call. Unless we up the abstraction for this controller and actually pass in the `id` the element is looking for under `observe(idRef)` so the controller callback is pre-filtered for the component use.

src/components/dialog/dialog.ts
```diff
@@ -80,92 +99,96 @@ export default class IgcDialogComponent extends EventEmitterMixin<
...
-   * Whether the dialog is opened.
-   * @attr
+   * Whether the dialog is open.
+   *
+   * Setting this property programmatically will open or close the dialog
+   * without animation and without emitting `igcClosing` / `igcClosed`.
+   * Prefer the `show()`, `hide()`, and `toggle()` methods for animated
+   * transitions, and note that events are only emitted when the dialog is
+   * closed through user interaction (the default action button, outside click,
+   * or the Escape key).
```
damyanpetev:
```suggestion
   * without animation and without emitting close events.
   * Prefer the `show()`, `hide()`, and `toggle()` methods for animated
   * transitions.
```

src/components/dialog/dialog.ts
```diff
@@ -80,92 +99,96 @@ export default class IgcDialogComponent extends EventEmitterMixin<
...
    * @attr keep-open-on-escape
+   * @default false
    */
   @property({ type: Boolean, attribute: 'keep-open-on-escape' })
   public keepOpenOnEscape = false;
 
   /**
-   * Whether the dialog should be closed when clicking outside of it.
+   * When set, clicking on the backdrop area outside the dialog surface
+   * will close it (emitting `igcClosing` / `igcClosed`).
```
damyanpetev:
```suggestion
   * When set, clicking on the backdrop area outside the dialog surface
   * will close it (emitting close events).
```

src/components/common/controllers/id-resolver.ts
rkaraivanov: It should since the constructor is always called in SSR environment. In that case the `_observer` will be undefined and the optional chaining will prevent exceptions.

damyanpetev (approved):

PR #2247, opened 2026-06-09
fix(slider): remove aria-label from host, leave only at thumb

### LINKED ISSUE #1410: [Bug]: Incorrect aria-label implementation on igc-slider component

didimmova:
Closes https://github.com/IgniteUI/igniteui-webcomponents/issues/1410#event-26459292868


simeonoff:
I'm closing this PR as the solution requires a more thorough refactoring of the entire component, and not just the thumb.

PR #2270, opened 2026-06-29
fix(skills): add LitElement shadow DOM grid theme injection pattern

### LINKED ISSUE #2271: [Bug]: igniteui-webcomponents-grids theme breaks inside LitElement shadow DOM

Other/unknown (please specify in the description)

### Description

Two independent agent runs building an igc-grid view inside a LitElement component hit different-looking bugs that share the same root: igniteui-webcomponents-grids is an Angular Elements bundle, and Angular Elements make assumptions about the DOM that break inside a shadow root.

### Reproduction

Create a LitElement component
Import the grid theme as a bare CSS import:
`import 'igniteui-webcomponents-grids/grids/themes/light/material.css';`
Render <igc-grid> with <igc-grid-toolbar> and <igc-grid-toolbar-hiding> inside the component's render()
Open the column chooser panel

### Workaround

No

### Is this a regression?

### Affected versions

latest

n/a

onlyexeption:
Closes #2271 

## Description

> *Provide a clear and concise summary of the changes in this PR and the motivation behind them*

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (code improvements without functional changes)

## Related Issues

Closes #

## Checklist

- [ ] My code follows the project's coding standards
- [ ] I have tested my changes locally
- [ ] I have updated documentation if needed
- [ ] Breaking changes are documented in the description


skills/igniteui-wc-generate-from-image-design/references/gotchas.md
```diff
@@ -134,6 +134,25 @@ For core UI component theming, prefer `create_component_theme` and apply the ret
 ### Grid theming is package-specific
 `igniteui-grid-lite` and the premium grid packages do not map to Angular's `igx-grid__*` internal class structure. Use `get_component_design_tokens("grid")`, the exact grid package docs, and exposed tokens or parts for the package present in the workspace.
 
+### `igniteui-webcomponents-grids` theme must be injected into the shadow root
+
+A bare CSS import (`import '...material.css'`) lands in the document head and is blocked by a LitElement shadow root — it never reaches the grid's internal `igx-*` components (toolbar checkboxes, column-actions panels, dropdown overlays). Without it those elements render with broken dimensions; checkboxes can expand to 1100px wide and collapse the entire grid UI.
```
damyanpetev: > A bare CSS import (`import '...material.css'`) lands in the document head

\* most commonly, based on your bundling tool, e.g. Vite, and being part of the "light" DOM, selector rules do not apply any element inside the Shadow DOM - the internal **_structure_** of the grid - igx components or not (so no relevance to this skill), which relies on the theme.

```suggestion
A bare CSS import (`import '...material.css'`) lands in the document head and selector rules outside the Shadow root never reach the grid and its internal structure. Without it those elements render with broken dimensions; checkboxes can expand to 1100px wide and collapse the entire grid UI.
```
^ I tried, can be worded better prob.

IIRC, you can also `<link rel="stylesheet" href="...material.css"` depending on asset handling in the app and let the browser caching handle it. I think we achieved that multiple ways.

damyanpetev: Also, hopefully we always enforce the `igniteui-webcompoents` theme on the document root too, because while the grid inside Shadow DOM need the theme inside for structural rules, it also very much needs the `:root` CSS variables, which do not apply from inside the Shadow root and much be outside. So if the grid is the only component, theme must be added twice :D

That's ofc a temporary situation, since we're working on bunding the styling with the component itself (like web components in this repo do) and the need for this will be removed almost completely after that.

damyanpetev: But the more important part is this is all dependent on the components in being inside Shadow DOM, and ideally we should guide apps _not_ to use that for view building - outside of building custom components for the app internally - the views themselves don't need it. That was purely for CSS isolation purposes, which can be managed on an app level _or_ using CSS modules like the React projects we have do and the views no longer need it to be shadow. Think we were meant to do that to the WC project too.

// 2c over

skills/igniteui-wc-integrate-with-framework/SKILL.md
```diff
@@ -90,13 +90,14 @@ Before loading a reference, identify the target framework from the project conte
 
 ---
 
-### 4. Vanilla JavaScript / HTML (fallback)
+### 4. Vanilla JavaScript / HTML / LitElement (fallback)
```
damyanpetev: Wouldn't really call Lit vanialla, but sure

kdinev (dismissed):

kdinev (approved):

PR #2276, opened 2026-06-30
chore: replace private jsdoc tags

damyanpetev:
The `@private` tag is not part of [TSDoc](https://tsdoc.org) and [TypeDoc doesn't recommend its usage](https://typedoc.org/documents/Tags._private.html), plus I find it redundant in TypeScript environment anyway.

Assuming this doesn't affect much else, but check CEM perhaps.

Tempted to slap `@internal` on `EventEmitterInterface.emitEvent` as well, since that's mostly for us to dispatch events internally and has no reason for being user-facing.
Also strongly suggest turning [`stripInternal`](https://www.typescriptlang.org/tsconfig/#stripInternal) if not on already :)

### DIFF
--- a/src/components/tree/tree.ts
+++ b/src/components/tree/tree.ts
@@ -55,10 +55,10 @@ export default class IgcTreeComponent extends EventEmitterMixin<
     }
   );
 
-  /** @private @hidden @internal */
+  /** @hidden @internal */
   public selectionService!: IgcTreeSelectionService;
 
-  /** @private @hidden @internal */
+  /** @hidden @internal */
   public navService!: IgcTreeNavigationService;
 
   /**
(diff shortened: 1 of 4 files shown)


src/components/tree/tree.ts
```diff
@@ -181,7 +181,7 @@ export default class IgcTreeComponent extends EventEmitterMixin<
   }
 
   /* blazorSuppress */
-  /** @private */
+  /** @hidden @internal */
   public expandToItem(item: IgcTreeItemComponent): void {
```
damyanpetev, line 185: This seems a bit odd to me - there seems to be no API equivalent to this, both the tree item and tree have `expand()`/`collapse()` which act on the singular item(s) and don't affect the parent structure, even the tree method. I kept this since neither the [spec](https://github.com/IgniteUI/igniteui-webcomponents/wiki/Tree-Specification#API) or [docs](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/tree) mention programmatic expansion to specific hierarchy.

However, as the docs themselves are an example, such functionality seems quite common place and we might want to consider something built-in for such use.

rkaraivanov: Looking at the usage, it is an internal method invoked by the tree nodes when their active state is changed.

rkaraivanov (approved):

PR #2283, opened 2026-07-02
refactor(tree): rewrite internals for performance, fix scroll/animaton regressions

rkaraivanov:
## Description

Rearchitect igc-tree/igc-tree-item's internal wiring while preserving the public API (properties, attributes, events, methods, CSS parts/custom properties) and existing behaviors.

- Replace `closest('igc-tree')` DOM-traversal wiring with a `@lit/context` provider/consumer (new tree.context.ts) so items resolve their owning tree synchronously without querying the DOM.
- Replace `querySelectorAll`-based child/descendant resolution with native light-DOM `.children` traversal; drop support for tree items wrapped in intermediate elements (e.g. `<div>`) in favor of direct nesting.
- Rewrite keyboard navigation on top of the shared `key-bindings` controller instead of a manual keydown switch statement; compute the navigable/visible item set on demand instead of eagerly caching it, removing the O(n²) recompute-on-every-mount root cause of prior sluggishness.
- Fix a real algorithmic issue in child/descendant collection: replace recursive `flatMap` + spread concatenation (quadratic for deeply nested trees) with a single-pass accumulator.
- Add `contain: layout` to the tree container and each tree-item host to scope reflow cost, smoothing out the expand/collapse height animation without reintroducing the compositing/scroll-jank risk that ruled out `content-visibility` for this component.
- Fix keyboard-navigation scroll lag: focus-driven scrollIntoView now scrolls instantly instead of `smooth`, so holding an arrow key no longer causes the visible scroll position to fall behind the actual focused/active item.
- Normalize all private members in tree-item.ts to the project's `_`-prefix convention.

## Type of Change
- [x] Refactoring (code improvements without functional changes)

## Testing

Test suite: adapted tree-utils.spec.ts/tree.spec.ts fixtures for the dropped wrapper-element support; all other specs unchanged and passing against the new implementation (88/88).

## Checklist

### DIFF
--- a/src/components/tree/themes/container.base.scss
+++ b/src/components/tree/themes/container.base.scss
@@ -10,4 +10,5 @@ $theme: $bootstrap;
     display: block;
     overflow: auto;
     font-family: var(--ig-font-family);
+    contain: layout style;
 }
--- a/src/components/tree/themes/item.base.scss
+++ b/src/components/tree/themes/item.base.scss
@@ -6,6 +6,7 @@
     display: flex;
     flex-direction: column;
     font-family: inherit;
+    contain: layout;
 
     igc-circular-progress {
         --diameter: #{sizable(rem(18px), rem(20px), rem(24px))};
@@ -18,6 +19,7 @@
 
 [role~='group'] {
     overflow: hidden;
+    contain: layout;
 }
 
 :host(:not([expanded])) [role~='group'] {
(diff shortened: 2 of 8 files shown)


ChronosSF (approved):

ChronosSF:
@rkaraivanov , LGTM but I'll let you decide if you merge or wait for another reviewer (I primarily tested it)

Issue #2328, opened 2026-08-14
[Bug]: adoptRootStyles misses styles injected by custom renderers


simeonoff:
### Which component(s) are affected?

Chat

Theming & Styles

### Description

When `adoptRootStyles` is enabled, an Angular custom renderer can return a component whose global styles are injected into the document only when that renderer creates its embedded view. The returned component is instantiated correctly inside the Chat message Shadow DOM, but its structural styles are missing from that Shadow Root.

#### Expected behavior

Styles available in `document.styleSheets` after the custom renderer runs should be adopted into the renderer's Shadow Root. In the reproduction, the Angular avatar should render as a styled 40 × 40 avatar.

#### Actual behavior

The avatar element and its host classes are present, but it computes to `display: inline` with a size of `0 × 0`. Its structural stylesheet exists in `document.styleSheets`, but is absent from the `igc-chat-message` Shadow Root's `adoptedStyleSheets`.

#### Root cause

`IgcChatMessageComponent.update()` calls `shouldAdoptStyles()` before `super.update()`. `AdoptedStylesController` therefore clones and caches `document.styleSheets` before the custom renderer executes. The renderer then creates the Angular embedded view, which causes Angular to inject the component stylesheet into the document. Because the per-document stylesheet snapshot is already cached, the newly added stylesheet is never adopted.

This is not specific to Avatar. It can affect any framework/component that injects styles during renderer execution. The same controller is also consumed by Grid Lite for custom header and cell templates, so cache refresh behavior should account for those consumers as well.

### Reproduction

A reproduction is available on the Ignite UI for Angular scoped-styles branch:

https://github.com/IgniteUI/igniteui-angular/tree/simeonoff/scoped-styles

1. Clone `IgniteUI/igniteui-angular` and check out `simeonoff/scoped-styles`.
2. Install dependencies with `npm install`.
3. In `src/app/chat/chat.sample.html`, remove this direct Chat child:

   ```html
   <igx-avatar slot="prefix" [src]="senderAvatars['support']" shape="circle" size="small"></igx-avatar>
   ```

   This direct instance currently preloads the avatar stylesheet and masks the issue.

4. Run `npm start` and open `http://localhost:4200/chat`.
5. Inspect the `igx-avatar` rendered by `#headerRenderer` inside an `igc-chat-message` Shadow Root.

The sample already has:

```ts
public options = signal<IgxChatOptions>({
    // ...
    adoptRootStyles: true
});
```

[...]

PR #2331, opened 2026-08-14
fix(chat): adopt document styles injected after the initial adoption (#2328)

rkaraivanov:
## Description

The document stylesheets were cloned once and cached forever, so styles a custom renderer injects while creating its view - Angular component styles, for instance - never reached the shadow root and their content rendered unstyled.

- Track the document while the styles are adopted and reflect additions, removals and links that finish loading later in the adopting shadow roots
- Memoize the clones per source stylesheet so incremental injections don't re-clone the whole document
- Keep the theme styles, which the adoption dropped by rebuilding the stylesheet list from the component's static styles alone
- Re-adopt on reconnect instead of staying cleared until an unrelated property changed

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Checklist


simeonoff:
I tested a build from this branch in the angular repo with the Dev Demos - everything works correctly now.

simeonoff (dismissed):

rkaraivanov:
@simeonoff 
Can you re-test with the latest changes?

simeonoff:
> @simeonoff Can you re-test with the latest changes?

Still works just fine.

PR #1893, opened 2025-10-14
Chat component - speech to text

ivanvpetrov:
Add a speech to text functionality to the chat component. 

The actual transcription is done by one of two options: 
1. (entirely on frontend) Webspeech API 
https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API
3. (on complimentary backend) Backend handling the communication to a 3rd party provider : Google Speech/Vertex AI/etc
https://github.com/IgniteUI/igniteui-speech-to-text-server

PR #2380, opened 2026-09-14
feat(speech-to-text): add igc-speech-to-text component and providers

rkaraivanov:
## Description

Add a button component that turns speech into text through a SpeechToTextProvider. The built-in WebSpeechProvider wraps the Web Speech API and expands region-less language tags for Edge. The extras entry point ships WebSocketSpeechToTextProvider, which streams MediaRecorder audio to a WebSocket endpoint over a JSON + binary protocol documented in src/extras/speech-to-text-websocket-protocol.md.

Recognized text arrives through igcInterim and igcResult, sessions end with igcEnd, and errors use Web Speech API compatible codes. Firefox keeps the API behind media.webspeech.recognition.enable, so the component renders disabled there by default.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Checklist


ivanvpetrov:
This standalone implementation of the speech-to-text functionality is a strong improvement over the earlier Chat-coupled approach in #1893.

It is reusable beyond Chat (input, text-area, etc.), avoids a SignalR dependency, and still support both the browser Web Speech API and custom backend providers through the WebSocket.

One tradeoff is that it has to be manually integrated in a chart component in order to be used there, which is normal and not a problem. 

The only thing it lacks, compared with the earlier implementation, is the ability to auto-submit upon silence in the chat component. It's currently not possible, because the default submit operation is private `_sendMessage()`. For this to work there should be a small publc  `sendMessage()` method in the `IgcChatComponent`

@rkaraivanov Since I'll be closing my previous PR about the speech-to-text. Do you want any assistance on this one, besides reviewing it?

rkaraivanov:
@ivanvpetrov 
A review would be really helpful but maybe you can take a build of the PR and try the web socket provider with a backend service. Aside from testing correct functionality I am interested if the [protocol description](https://github.com/IgniteUI/igniteui-webcomponents/blob/8b6aaa767f1dd6d16023b8f0bb8e7ecdce36deb2/src/extras/speech-to-text-websocket-protocol.md) is comprehensive enough.

ivanvpetrov:
I've built an .NET backend server to test the WebSocket protocol and communication. I've set it up to route the transcription to Google.Speech.V2 since I'm already familiar with it.

The whole round trip is working properly. The only issue I had was with language negotiation, I had to hard code it to 'en-US'. So maybe we have a possible improvement point there. 
I'll continue analyzing and testing it and I'll let you know.

rkaraivanov:
> The whole round trip is working properly. The only issue I had was with language negotiation, I had to hard code it to 'en-US'. So maybe we have a possible improvement point there. I'll continue analyzing and testing it and I'll let you know.

@ivanvpetrov 
Was that on the client side? Because there were issues with the full BCP 47 tags as seen here:
https://github.com/IgniteUI/igniteui-webcomponents/blob/8b6aaa767f1dd6d16023b8f0bb8e7ecdce36deb2/src/components/speech-to-text/providers/web-speech.ts#L62-L81

